PR #3, opened 2020-05-20
Update image improvements

themylogin:


scripts/build.sh
```diff
@@ -478,9 +490,8 @@ update_git_repo() {
 	GHBRANCH="$2"
 	REPO="$3"
 	echo "`date`: Updating git repo [${NAME}] (${LOG_DIR}/git-checkout.log)"
-	(cd ${SOURCES}/${NAME} && git reset --hard) >${LOG_DIR}/git-checkout.log 2>&1 || exit_err "Failed git reset"
-	(cd ${SOURCES}/${NAME} && git fetch --unshallow) >${LOG_DIR}/git-checkout.log 2>&1
-	(cd ${SOURCES}/${NAME} && git pull origin ${GHBRANCH}) >${LOG_DIR}/git-checkout.log 2>&1 || exit_err "Failed git pull"
+	(cd ${SOURCES}/${NAME} && git fetch origin ${GHBRANCH}) >${LOG_DIR}/git-checkout.log 2>&1 || exit_err "Failed git fetch"
```
kmoore134, line 493: Does this succeed with the initial repos being checked out shallow? When I tested originally I couldn't get it to work again without running --unshallow first...

truenas_install/__main__.py
```diff
@@ -0,0 +1,47 @@
...
+import logging
+import subprocess
+import sys
+
+logger = logging.getLogger(__name__)
+
+if __name__ == "__main__":
+    update = json.loads(sys.stdin.read())
+
+    root = update["root"]
```
william-gr, line 13: I wonder if we shouldn't create the new ZFS dataset from here so its unified between install and update?

truenas_install/__main__.py
```diff
@@ -0,0 +1,47 @@
...
+
+    undo = []
+    try:
+        subprocess.run(["mount", "-t", "proc", "none", f"{root}/proc"], check=True)
+        undo.append(["umount", f"{root}/proc"])
+
+        subprocess.run(["mount", "-t", "sysfs", "none", f"{root}/sys"], check=True)
+        undo.append(["umount", f"{root}/sys"])
+
+        for device in glob.glob("/dev/sd?*") + ["/dev/zfs"]:
```
william-gr, line 34: It could also be vd*

Wonder if there are others as well

scripts/build_manifest.py
```diff
@@ -0,0 +1,31 @@
...
+        for file in files:
+            abspath = os.path.join(root, file)
+            checksums[os.path.relpath(abspath, output)] = subprocess.run(
+                ["sha1sum", abspath],
+                check=True, stdout=subprocess.PIPE, encoding="utf-8", errors="ignore",
+            ).stdout.split()[0]
+
+    with open(os.path.join(output, "manifest.json"), "w") as f:
+        f.write(json.dumps({
+            "version": version,
```
william-gr, line 28: We lost date key? On purpose?

truenas_install/__main__.py
kmoore134: Here's list we currently have: 
https://github.com/truenas/truenas-installer/blob/master/usr/sbin/truenas-install#L139

scripts/build_manifest.py
themylogin: Done

scripts/build.sh
themylogin: Fixed it

william-gr:
Good job!

PR #10, opened 2020-06-08
Write required disk size to update manifest

themylogin:


scripts/build_manifest.py
```diff
@@ -29,5 +29,7 @@
             "date": datetime.utcnow().isoformat(),
             "version": version,
             "checksums": checksums,
+            # Unpacked root squashfs consumes on average twice its size on ZFS dataset + 10% just in case
+            "size": int(os.path.getsize(os.path.join("output", "rootfs.squashfs")) * 2 * 1.1),
```
william-gr, line 33: Should the size just reflect reality and we put the 2 * 1.1 in the code?

themylogin: I prefer to leave more control in our (update file providers) hands, e.g. we can calculate this value precisely if we ever need this.

william-gr: I dont understand, how that does not leave the control in our hands? The size is a property in the manifest, why purpose does it serve the manfiest to lie about the exact size of a file we ship?

themylogin: We say "to install this, you'll need at least `size` space available in your boot pool"

william-gr: Thats a guess at best. I think we need to either save the real size it uses before squashing it if thats what we are after.

PR #14, opened 2020-06-08
Upload packages immediately after build so one can do apt-get update  faster

themylogin:


kmoore134:
Only concern is that by pushing packages before the "update / iso" steps, is that if there is an issue with any of the packages built, we won't know until after they've already landed on APT server. I.E. if verify_rootfs fails, or package has bad depends that prevent it from building the update image, etc.

If that is OK, then I don't see any problem with this change.

themylogin:
It's ok for me, we only use this repo internally. Thoughts, @william-gr ?

PR #19, opened 2020-06-16
Github Action ubuntu build

### LINKED ISSUE #16: [Feature Request] GitHub Actions
We could implement a GitHub action that will run the build process on pushes/pr using a prestine Debian image for CI purposes to validate that no changes have broken the build process. If you wish this is something I can implement and can be expanded if need be later. The action will be for CI and not CD so the images it builds will not be accessable but it is an option.

luigi311:
This will add a build test in ubuntu 18.04 to validate that no changes have broken the build. This should allow for easy expanding to more OS by just adding more lines to the include section on line 19, and adding their method for installing dependencies if apt is not applicable. This is set to run on both push and pull requests.

Closes #16

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,39 @@
+name: build
+on:
+  push:
+    paths-ignore:
+      - .gitignore
+      - "*.md"
+  pull_request:
+    paths-ignore:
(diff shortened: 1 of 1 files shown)


william-gr:
Thanks for your contribution

PR #22, opened 2020-06-20
Enable user enabled services when user updates

sonicaj:
To give some context, when we create a new BE, the specified defaults we have set for services are to disable them and a new BE has the vendor configured settings ( the preset file we ship ) being applied. So any services which the user had enabled in his old BE won't boot in the newer BE as this is remembered by systemd which is entirely replaced in the newer BE.
In this case we need to enable the services user had explicitly marked as enabled in the new BE so when the user boots, he/she has those services running.

### DIFF
--- a/truenas_install/__main__.py
+++ b/truenas_install/__main__.py
@@ -51,6 +51,17 @@ def run_command(cmd, **kwargs):
         raise
 
 
+def enable_user_services(root):
+    from middlewared.client import Client
+    with Client() as cl:
+        systemd_units = []
+        for srv in cl.call('service.query', [['enable', '=', True]]):
+            systemd_units.extend(cl.call('service.systemd_units', srv['service']))
+
+    if systemd_units:
+        run_command(['systemctl', f'--root={root}', 'enable'] + systemd_units)
+
+
 if __name__ == "__main__":
     input = json.loads(sys.stdin.read())
 
@@ -127,6 +138,8 @@ if __name__ == "__main__":
 
                     with open(f"{root}/data/need-update", "w"):
                         pass
+
+                    enable_user_services(root)
                 else:
                     run_command(["cp", "/etc/hostid", f"{root}/etc/"])
 
--- a/truenas_install/__main__.py
+++ b/truenas_install/__main__.py
@@ -5,6 +5,7 @@ import json
 import logging
 import os
 import re
+import sqlite3
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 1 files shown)


sonicaj:
> Won't middleware enable/disable these services when booting in new BE before doing anything?

When middleware enables/disables services - systemd has already started the boot process and we are unable to change the order at that point resulting in the services not starting for the first time after update.

themylogin:
@sonicaj same will happen with config import, so that's not a solution. I think we'll have to start those services on boot manually if they were disabled in systemd.

themylogin:
Also we'll have to stop disabled services.

sonicaj:
That does not work unfortunately as systemd does not respect that file like that.

themylogin:
@sonicaj why? Does it modify `10-truenas_user.preset` in-place?

sonicaj:
I am not sure I follow. Presets are generally for vendors to ship software with knobs configured to enable/disable services as reasonable defaults. Just adding a user preset file does not change that behaviour. 
In technical terms, this file just lets us know what units are enabled disabled, it serves no other purpose. We can use it to force systemd to fallback to specified defaults but that’s not the same as disabling/enabling the service which we do here.

themylogin:
> In technical terms, this file just lets us know what units are enabled disabled,

Sorry, I did not understand this at first. Then I don't think we need this file at all, we already have a database file from the old BE. Let's use sqlite code from https://github.com/freenas/freenas/pull/4991/files to read it and enable/disable units. Also I think chrooting might do a better job than `systemd --root`

sonicaj:
That will not be enough unfortunately. Why is that is because the services we define in database may have one or more or different unit files then the name of the service stored in database. 

So we don’t have the information to enable/disable systemd units correctly.

truenas_install/__main__.py
```diff
@@ -51,6 +51,20 @@ def run_command(cmd, **kwargs):
         raise
 
 
+def enable_user_services(root, old_root):
+    user_services_file = os.path.join(old_root or '/', 'data/user-services.json')
```
themylogin, line 55: `old_root` is always set here

truenas_install/__main__.py
```diff
@@ -51,6 +51,20 @@ def run_command(cmd, **kwargs):
...
+    if not os.path.exists(user_services_file):
+        return
+
+    with open(user_services_file, 'r') as f:
+        systemd_units = [
+            srv for srv, enabled in json.loads(f.read()).items() if enabled
+        ]
+
+    if systemd_units:
+        run_command(['systemctl', f'--root={root}', 'enable'] + systemd_units)
```
themylogin, line 65: Use `chroot` please just in case

PR #24, opened 2020-06-24
Add default syslog-ng config

themylogin:


themylogin:
@william-gr it is responsible for writing text logs (journald only writes it's own binary format) and also sending to remote syslog.

sonicaj:
@themylogin shouldn't we be doing this in a single place ? We are already copying files for mail in truenas post install. If you think this is a better place, we can do that here ?

themylogin:
I agree with you, thank you for noticing this

themylogin:
Are we sure `truenas` script is the last thing that gets executed in the package installation stage? (I need this for groups and passwd)

sonicaj:
@themylogin we should make sure if it isn't. To my understanding, it acts a meta port finalising the dependencies, so it's post install should be good for this.

PR #31, opened 2020-07-14
paramiko

themylogin:


scripts/build.sh
```diff
@@ -406,8 +406,12 @@ build_dpkg() {
 	fi
 
 	# Make a programatically generated version for this build
-	DATESTAMP=$(date +%Y%m%d%H%M%S)
-	chroot ${DPKG_OVERLAY} /bin/bash -c "cd $srcdir && dch -b -M -v ${DATESTAMP}~truenas+1 --force-distribution --distribution bullseye-truenas-unstable 'Tagged from truenas-build'" || exit_err "Failed dch changelog"
+	if [ "$name" -ne "paramiko" ] ; then
```
william-gr, line 409: Hard-coding a package name here? Why? Thats sounds wrong :P

themylogin: Unfortunately I am not very familiar with the strange programming language this code is written in 😞

william-gr: Why do we need not to change paramiko version?

Alternatively we could have a flag in the manifest to not do that?

Thoughts @kmoore134 ?

themylogin: Otherwise `dch` complains and renames source directory, and I am not able to chdir there via `subdir` parameter anymore

kmoore134: The update you did to make a generic "generate_version" looks like how I would have solved it. Abstracted away so there is no hard-coded name, and we can re-use it if other packages require it later.

PR #35, opened 2020-08-05
Add nvme detection

sionicion:
The truenas SCALE installer was failing in part because of this code when installing to an nvme disk. I updated the code to check if nvme or non-nvme and use the correct naming convention accordingly.

### DIFF
--- a/truenas_install/__main__.py
+++ b/truenas_install/__main__.py
@@ -184,9 +184,13 @@ if __name__ == "__main__":
                     os.makedirs(f"{root}/boot/efi", exist_ok=True)
                     for disk in disks:
                         run_command(["chroot", root, "grub-install", "--target=i386-pc", f"/dev/{disk}"])
-
-                        run_command(["chroot", root, "mkdosfs", "-F", "32", "-s", "1", "-n", "EFI", f"/dev/{disk}2"])
-                        run_command(["chroot", root, "mount", "-t", "vfat", f"/dev/{disk}2", "/boot/efi"])
+                        # Check if nvme disk and instead use nvme naming convention.
+                        if ("nvme" in disk):
+                            run_command(["chroot", root, "mkdosfs", "-F", "32", "-s", "1", "-n", "EFI", f"/dev/{disk}p2"])
+                            run_command(["chroot", root, "mount", "-t", "vfat", f"/dev/{disk}p2", "/boot/efi"])
+                        else:
+                            run_command(["chroot", root, "mkdosfs", "-F", "32", "-s", "1", "-n", "EFI", f"/dev/{disk}2"])
+                            run_command(["chroot", root, "mount", "-t", "vfat", f"/dev/{disk}2", "/boot/efi"])
                         try:
                             run_command(["chroot", root, "grub-install", "--target=x86_64-efi",
                                          "--efi-directory=/boot/efi",
--- a/truenas_install/__main__.py
+++ b/truenas_install/__main__.py
@@ -185,12 +185,10 @@ if __name__ == "__main__":
                     for disk in disks:
                         run_command(["chroot", root, "grub-install", "--target=i386-pc", f"/dev/{disk}"])
                         # Check if nvme disk and instead use nvme naming convention.
-                        if ("nvme" in disk):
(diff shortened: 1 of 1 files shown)


truenas_install/__main__.py
```diff
@@ -184,9 +184,13 @@ def enable_user_services(root, old_root):
                     os.makedirs(f"{root}/boot/efi", exist_ok=True)
                     for disk in disks:
                         run_command(["chroot", root, "grub-install", "--target=i386-pc", f"/dev/{disk}"])
-
-                        run_command(["chroot", root, "mkdosfs", "-F", "32", "-s", "1", "-n", "EFI", f"/dev/{disk}2"])
-                        run_command(["chroot", root, "mount", "-t", "vfat", f"/dev/{disk}2", "/boot/efi"])
+                        # Check if nvme disk and instead use nvme naming convention.
+                        if ("nvme" in disk):
+                            run_command(["chroot", root, "mkdosfs", "-F", "32", "-s", "1", "-n", "EFI", f"/dev/{disk}p2"])
```
themylogin, line 189: Let's try to avoid code duplication and only calculate `partition` variable depending on whether if disk is or if not an NVMe

truenas_install/__main__.py
```diff
@@ -184,9 +184,13 @@ def enable_user_services(root, old_root):
                     os.makedirs(f"{root}/boot/efi", exist_ok=True)
                     for disk in disks:
                         run_command(["chroot", root, "grub-install", "--target=i386-pc", f"/dev/{disk}"])
-
-                        run_command(["chroot", root, "mkdosfs", "-F", "32", "-s", "1", "-n", "EFI", f"/dev/{disk}2"])
-                        run_command(["chroot", root, "mount", "-t", "vfat", f"/dev/{disk}2", "/boot/efi"])
+                        # Check if nvme disk and instead use nvme naming convention.
+                        if ("nvme" in disk):
```
themylogin, line 188: `(` and `)` are redundant here

sionicion: `                        # Check if nvme disk and instead use nvme naming convention.
                        if "nvme" in disk: partition = "{disk}p2"
                        else: partition = "{disk}2"
                        run_command(["chroot", root, "mkdosfs", "-F", "32", "-s", "1", "-n", "EFI", f"/dev/{partition}"])
                        run_command(["chroot", root, "mount", "-t", "vfat", f"/dev/{partition}", "/boot/efi"])`

Is this better?

themylogin: Yes!

kmoore134:
ok to test

truenas_install/__main__.py
```diff
@@ -184,9 +184,11 @@ def enable_user_services(root, old_root):
                     os.makedirs(f"{root}/boot/efi", exist_ok=True)
                     for disk in disks:
                         run_command(["chroot", root, "grub-install", "--target=i386-pc", f"/dev/{disk}"])
-
-                        run_command(["chroot", root, "mkdosfs", "-F", "32", "-s", "1", "-n", "EFI", f"/dev/{disk}2"])
-                        run_command(["chroot", root, "mount", "-t", "vfat", f"/dev/{disk}2", "/boot/efi"])
+                        # Check if nvme disk and instead use nvme naming convention.
+                        if "nvme" in disk: partition = "{disk}p2"
```
themylogin, line 188: The correct PEP8 syntax is:
```
if "nvme" in disk:
    partition = "{disk}p2"
else:
    partition = "{disk}2"
```

themylogin: Also you've forgot to make them f-strings e.g. `f"{disk}2"`

sionicion: Ah I just saw this after I committed more changes, I honestly never touched Python in my life so I'm winging it. I discovered the issue with the f-strings after it failed for me on my VM. I'll recommit with the PEP8 syntax now.

sionicion:
Ok so the latest ISO I built is working correctly with this code for both nvme and non-nvme drives. Anything else need done?

themylogin:
@sionicion now looks good to me, thank you!

PR #41, opened 2020-08-24
NAS-107334 \ Build ZFS with QAT support

Ornias1993:
Currently SCALE only builds the ZFS kernel module without QAT support.
This PR builds ZFS with QAT support baked-in.

It does not add the QAT drivers to the SCALE build, just makes sure ZFS is compatible with QAT. This way experienced users could install the drivers themselves and get QAT running without needing to rebuild ZFS

TODO:
- Include QAT driver/software with TrueNAS (in later PR or never)

Notes:
- While I don't expect QAT to be officially supported by IX, at least including it in the build makes it easier for people to enable it manually (without rebuilding ZFS)
- I don't think there is a nice way of adding the actual QAT driver to SCALE... No idea honestly. So this is just a patch for having ZFS support QAT so people don't have to rebuild ZFS to get WAT support

Signed-off-by: Kjeld Schouten-Lebbing <kjeld@schouten-lebbing.nl>

### DIFF
--- a/scripts/mk-zfs-modules
+++ b/scripts/mk-zfs-modules
@@ -5,10 +5,31 @@ db_set zfs-dkms/note-incompatible-licenses "true"
 
 #KERNEL_VERSION=4.19.0-5-amd64 # could be a Jenkins parameter
 KERNEL_VERSION=$(dpkg --get-selections | grep -e "^linux-image-[0-9]" | awk '{print $1}' | sed 's|linux-image-||g')
+export KERNEL_SOURCE_ROOT=/lib/modules/${KERNEL_VERSION}/build
 
 apt -qyy install \
 	linux-image-${KERNEL_VERSION} \
-	linux-headers-${KERNEL_VERSION}
+	linux-headers-${KERNEL_VERSION} \
+	pciutils \
+	wget \
+	pkg-config \
+	git \
+	libudev-dev
+	
+# Setup QAT driver
+mkdir -p /QAT/FETCH
+wget https://01.org/sites/default/files/downloads/qat1.7.l.4.10.0-00014.tar.gz -O /QAT/FETCH/qat.tar.gz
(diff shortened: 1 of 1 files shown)


Ornias1993:
Note: This PR is mostly waiting on new QAT drivers that natively support kernel 5.6+

kmoore134:
ok to test

Ornias1993:
@kmoore134 Still working on it, looks like the new drivers need a little more work still...
Will remove draft when "really" ready, thanks for the support though :)

Ornias1993:
With #51 this would basically have to be upstreamed I think, which isn't going to happen any time soon.
Unless @freqlabs has a good idea how to get it in, I think I can just as well close this.

freqlabs:
I imagine you could build QAT as a package like everything else and put it before ZFS in the list so it can be added as a dependency. It's not a priority for us to support QAT at this time, but that might at least get you something to use locally if it matters that much to you.

Ornias1993:
@freqlabs You don't have to repeat yourself over and over, I know it has close to no priority. I just was trying to help the people that want to use it and ask a friendly question about it. I find it somewhat passive agressive to bring that up yet again to be honest.

I've no idea how to package QAT as a package, because no one did so before me, i've no experience in it and intel doesn't support it either. So i'll close it, untill someone points out of to get the damn thing into a deb package.

freqlabs:
My apologies, no hostility was intended.

For getting a QAT deb to build, this might serve as a good example: https://github.com/truenas/intel-pcm

Ornias1993:
@freqlabs No hard feelings :)
Thank you very much, thats indeed something I can work with ^^

Ornias1993:
Okey, I'm quiting work on this.
Intel took about 6 months(!) to implement and release the (already widely known fixes) for intel QAT and kernel version 5.6.
Now their software breaks due to changes in kernel... 5.7 which will (no doubt) take another 6 months.

It would either mean we have to patch every QAT release ourselves to keep up with SCALE's kernel versioning or delay updating the SCALE kernel version every time. I think both are borderline unacceptable. So intel QAT is not an option, due to intel being slow.

PR #43, opened 2020-09-04
Add firmware-bnx2 package to TrueNAS build

dylanmtaylor:
This is required for ethernet to function on many Dell servers, including my R710

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -46,6 +46,7 @@
 	"base-packages": [
 		"dosfstools",
 		"consul",
+		"firmware-bnx2",
 		"grub-pc",
 		"grub-pc-bin",
 		"grub-efi-amd64",


dylanmtaylor:
I can install this myself by copying firmware-bnx2_20200721-1_all.deb to a flash drive and running dpkg -i on it, but it's not ideal. I'd love to have networking work out of the box on my server.

dylanmtaylor:
Immediately after running dpkg -i on the package, then ifconfig up on eno1 (the interface I have plugged in), I was able to go back to the menu, choose configure network interfaces, select the interface and I was able to get a connection after removing existing network configuration.

dylanmtaylor:
I'm also not the only one with this issue: https://www.ixsystems.com/community/threads/nic-driver-issues-with-install-on-dell-r410.86511/post-601464

kmoore134 (approved):

kmoore134:
ok to test

PR #45, opened 2020-09-04
Change Step Order to Fix ISO Upload

dylanmtaylor:
In my commit yesterday I didn't notice that the clean step deletes the ISO right before it can be uploaded. This minor PR moves the clean step to after the ISO is uploaded.


dylanmtaylor:
This is a working run of this change on my fork, with the ISO in the artifacts: https://github.com/dylanmtaylor/truenas-build/actions/runs/239289338

PR #50, opened 2020-09-18
python3-psutil

themylogin:


william-gr (commented):
Should that be a dependency of truenas-installer instead?

PR #52, opened 2020-09-21
[WIP] NAS-107334 \ Build ZFS with QAT support

Ornias1993:
DRAFT: This is more of a WIP playground to look into getting this running and is not supposed to work or be merged.

DRAFT DONE:
- Get QAT building on/for Kernel 5.8

DRAFT TODO:
- Get a manifest to compile source from repo on TrueNAS SCALE build
- Get the source into the ZFS buildscript
- Use the source to build ZFS with QAT support

Currently SCALE only builds the ZFS kernel module without QAT support.
This PR builds ZFS with QAT support baked-in.

It does not add the QAT drivers to the SCALE build, just makes sure ZFS is compatible with QAT. This way experienced users could install the drivers themselves and get QAT running without needing to rebuild ZFS

TODO:
- Include QAT driver/software with TrueNAS (in later PR or never)

Notes:
- While I don't expect QAT to be officially supported by IX, at least including it in the build makes it easier for people to enable it manually (without rebuilding ZFS)
- I don't think there is a nice way of adding the actual QAT driver to SCALE... No idea honestly. So this is just a patch for having ZFS support QAT so people don't have to rebuild ZFS to get WAT support

Signed-off-by: Kjeld Schouten-Lebbing <kjeld@schouten-lebbing.nl>

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -69,6 +69,7 @@
 		"nvidia-kernel-450.66",
 		"nvidia-container-toolkit",
 		"nvidia-smi",
+		"intel-qat",
 		"openzfs",
 		"libnvidia-encode1",
 		"squashfs-tools",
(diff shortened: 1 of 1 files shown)


Ornias1993:
Closing.
Considering the maintenance load it would create. IX would basically have to maintain the QAT drivers because Intel doesn't really maintain it and future QAT drivers will not be the same package (which means we can only support one driver version). 

The idea of supporting QAT is nice, in practice it turns a maintenance nightmare.

PR #69, opened 2020-10-16
Pyroute2

themylogin:


conf/build.manifest
```diff
@@ -294,6 +299,8 @@
 				"pkg-pybonjour",
 				"pydevd",
 				"python-netsnmpagent",
+				"python-pyroute2",
+				"python-remote-pdb",
```
william-gr, line 303: Do we need these two here? isnt in middlewared repo enough?

themylogin: Looks like we do, this package wasn't built during the last build

william-gr: So yes for python-pyroute2, but no for python-remote-pdb ?

themylogin: How is it different? The same thing will happen if we upgrade it, it will only be rebuild during full build.

william-gr: I missed the remote-pdb being a repo of our own

william-gr (approved):

PR #70, opened 2020-10-19
Add nss-pam-ldapd to build

anodos325:


william-gr (approved):

william-gr:
@themylogin do we also have to add this package to dependencies? I dont recall

sonicaj:
@william-gr we don't have to. The idea was just to give better clarity to define the order in which the packages should be built and maybe at a later stage do some validation beforehand to ensure we have the necessary packages being built before a certain package.
Although i think it has created more confusion : )
Please advise if you think we should maybe remove this for now ?

PR #71, opened 2020-10-20
NAS-108003 / 20.12 / Add "waagent", helper agent which improves support when running SCALE

kmoore134:
virtualized inside Azure

https://packages.debian.org/bullseye/waagent


william-gr (approved):

kmoore134:
retest this please

PR #72, opened 2020-10-22
Build swagger package

sonicaj:


kmoore134 (approved):

william-gr (commented):
Vladimir said we need to add the package as a dependency otherwise incremental builds may not build it.

@themylogin ?

themylogin:
I was incorrect. `dependencies` key is not read in build script, we should probably just remove this key to avoid confusion.

william-gr (approved):

PR #73, opened 2020-10-24
Update debian python packages from PIP

themylogin:
@william-gr I don't like the idea of making 6 debian packages just for the sake of being able to close persistent HTTPS connection to GMail API when changing configuration. hope you'll understand me.

### DIFF
--- a/conf/requirements.txt
+++ b/conf/requirements.txt
@@ -1,0 +1,6 @@
+google-api-core==1.23.0
+google-api-python-client==1.12.5
+google-auth==1.22.1
+googleapis-common-protos==1.52.0
+protobuf==3.13.0
+uritemplate==3.0.1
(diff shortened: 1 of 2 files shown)


william-gr (changes requested):
I understand the intents, but I can't agree with this since it breaks all the package aspects from debian.

Same is true for FreeBSD, we have ports for a reason, its not different here.

I dont mind if we create some generic python repo where we put pip fetches all there in a single place, but we need one package per python module. We cant just change that because we are bored and want a little bit less of work.

themylogin:
@william-gr so the same thing but in a separate package will be acceptable?

william-gr:
We need one package per python module, they can all be in the same repo.

We will eventually start upstreaming what we have.

PR #85, opened 2020-12-14
NAS-108649 / 21.02 / nvme/mmc compatibility: look for partitions before using

dhiltonp:
When paired with https://github.com/truenas/truenas-installer/pull/24, installation should work on virtually all block devices. I have confirmed functionality by installing on an emmc drive.

### DIFF
--- a/truenas_install/__main__.py
+++ b/truenas_install/__main__.py
@@ -50,6 +50,12 @@ def run_command(cmd, **kwargs):
         write_error(f"Command {cmd} failed with exit code {e.returncode}: {e.stderr}")
         raise
 
+def get_partition(disk, partition):
+    paths =[f"/dev/{disk}{partition}", f"/dev/{disk}p{partition}"]
+    for path in paths:
+        if os.path.exists(path):
+            return path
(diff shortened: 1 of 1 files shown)


themylogin (approved):

truenas_install/__main__.py
```diff
@@ -111,7 +117,7 @@ def install_grub_freebsd(input, manifest, pool_name, dataset_name, disks):
             """
         else:
             efi_partition_uuid = run_command([
-                "grub-probe", "--device", f"/dev/{disks[0]}p1", "--target=fs_uuid"
+                "grub-probe", "--device", get_partition(disks[0], 1), "--target=fs_uuid"
```
dhiltonp, line 120: Since you're using a bug tracker and assigning priority: this line as originally written only allowed TrueNAS to install on EFI systems with NVME installation target drives.

This change also allows TrueNAS to be installed on EFI systems onto HDDs and SATA SSDs (including M.2 SATA drives).

themylogin (approved):

kmoore134 (approved):

PR #89, opened 2020-12-23
NAS-108790 / 21.02 / add xtail and create community/sales/support sections for pkgs

yocalebo:


william-gr (commented):
Not sure what I think about segregating these in code, as we can have mixed requests.
To me they are mostly comments not sections but I would like to hear others opinions

yocalebo:
> Not sure what I think about segregating these in code, as we can have mixed requests.
> To me they are mostly comments not sections but I would like to hear others opinions

Yeah, I agree. I'm not requiring this but it was the idea I had at the time. Would like the input from others as well. I still think we need to somehow segregate what was requested from the "community" as opposed to the "support/sales" team. It will, hopefully, allow us to prune old packages that we no longer need/aren't used etc.

PR #90, opened 2020-12-23
NAS-108788 / 21.02 / add xtail and create "additional-packages" section

yocalebo:


kmoore134:
My only concern would be that just because a package came in one door, it doesn't mean its invalid for the other use-cases. I.E. community may request something, but later support and sales start relying on it as well. Would it instead make more sense to just keep a package-changelog.md file or similar where we can update with comments when a new package is added? Explain where it came from, who requested, and then refer to that if we're auditing the list later? The issue with doing this in JSON is that no comments are allowed, so you don't really get any color, you can't link to a JIRA ticket, etc. Thoughts @william-gr ?

yocalebo:
> My only concern would be that just because a package came in one door, it doesn't mean its invalid for the other use-cases. I.E. community may request something, but later support and sales start relying on it as well. Would it instead make more sense to just keep a package-changelog.md file or similar where we can update with comments when a new package is added? Explain where it came from, who requested, and then refer to that if we're auditing the list later? The issue with doing this in JSON is that no comments are allowed, so you don't really get any color, you can't link to a JIRA ticket, etc. Thoughts @william-gr ?

We could extend the array structures that I've added and make them complex by adding a comment field? That way the JSON file would have the "comments" we just wouldn't use the information in it?

yocalebo:
i.e something like:

`"support-requested-packages":  [{"package": "xtail", "comment": "often used by support team"}]`

kmoore134:
I'd be OK with that compromise, so at least we can link to a ticket, or some short text by way of explanation.

yocalebo:
ping @kmoore134 @william-gr 😄

PR #95, opened 2021-01-19
Update aiohttp version

sonicaj:
We are seeing an issue after recent update of debian packages where aiohttp does not parse the URI correctly - however this is not reproducible on aiohttp 3.7.3 version which is the latest available and debian upstream just tagged it but it isn't available yet.

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -312,6 +312,11 @@
 			"branch": "main"
 		},
 		{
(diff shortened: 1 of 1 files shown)


yocalebo (approved):
Tested on my local SCALE cluster running a snapshot from 1-17-2021.

william-gr (approved):

PR #104, opened 2021-01-31
Remove repositories from build

sonicaj:
This PR adds changes to remove repos from build because some of them are not required anymore and some already have changed upstream to the version we desire so it's not necessary to maintain a fork or build them each time.

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -267,11 +267,6 @@
 			"branch": "master"
 		},
 		{
-			"name": "rclone",
-			"repo": "https://github.com/truenas/rclone",
-			"branch": "master"
-		},
-		{
 			"name": "zectl",
 			"repo": "https://github.com/truenas/zectl",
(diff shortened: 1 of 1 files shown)


william-gr:
My concerns are debootstrap and debian_archive_keyring, I think Kris had to change things in these, its not just about version. @kmoore134

sonicaj:
I consulted with @kmoore134 and he said that if apt works fine without these, we should be good. It did in my tests but i'll verify it again for good measure. ( we were only adding TN key in the repo which we are doing in truenas-build as well iirc )

william-gr (approved):

kmoore134:
Yea, this *should* be fine. Those original mods were done to work with the early builder framework, where packages came from mini-buildd, not locally built. Needs testing, but if Waqar has run it though its paces, it should be OK.

kmoore134 (approved):

PR #111, opened 2021-02-15
NAS-109351 / 21.04 / Generate native systemd unit files for SysV services that lack ones t…

themylogin:
…o prevent systemd-sysv-generator warnings

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -749,6 +749,19 @@ custom_rootfs_setup() {
 	echo 'ZFS_INITRD_POST_MODPROBE_SLEEP=15' >> ${CHROOT_BASEDIR}/etc/default/zfs
 	chroot ${CHROOT_BASEDIR} update-initramfs -k all -u
 
+	# Generate native systemd unit files for SysV services that lack ones to prevent systemd-sysv-generator
+	# warnings
+	mkdir ${CHROOT_BASEDIR}/tmp/systemd
+	chroot ${CHROOT_BASEDIR} /usr/lib/systemd/system-generators/systemd-sysv-generator \
+		/tmp/systemd /tmp/systemd /tmp/systemd
+	for file in ${CHROOT_BASEDIR}/tmp/systemd/*.service; do
+		echo >> $file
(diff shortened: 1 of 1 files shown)


yocalebo (commented):
Are you worried that the generated service files for a more complex service like `SCST`, for example, will not be correct? Thinking of the `RemainAfterExit` parameter and such.

themylogin:
@sonicaj it is known why scst does not have a native systemd unit file?

yocalebo:
> @sonicaj it is known why scst does not have a native systemd unit file?

I've looked into this and found https://github.com/bvanassche/scst/issues/29, as it's already been asked. I believe it's just a matter of not creating one upstream...

sonicaj:
@yocalebo yes, that's accurate. However i am not sure how this change changes anything wrt scst ?

yocalebo:
> @yocalebo yes, that's accurate. However i am not sure how this change changes anything wrt scst ?

Because automatically generating systemd unit files isn't perfect. `SCST` is a complex service so I'm wondering if the automatically generated one will be the "correct" one.

sonicaj:
Well it's already happening ? The concerns are valid but the scope of this change seems different ?

sonicaj:
I think we can perhaps create a new ticket to ensure that the services we officially support don't have an issue in their generated systemd files ( if they are generated ) and if they do, we can either create issues upstream and if possible maybe even contribute upstream. Thoughts ?

yocalebo:
> I think we can perhaps create a new ticket to ensure that the services we officially support don't have an issue in their generated systemd files ( if they are generated ) and if they do, we can either create issues upstream and if possible maybe even contribute upstream. Thoughts ?

If you think the generated service file works, then I'm fine and I agree, I only bring up `SCST` since I consider it a "core" service. Seems strange to not have a native systemd service file.

sonicaj:
@yocalebo to the best of my knowledge we have not had issues so far. I think we are good, and if we see issues we can of course fix those.

sonicaj (approved):

PR #120, opened 2021-03-01
Build with custom kernel

anodos325:
This adds linux kernel to the build. Due to size of repo and temporary files,
logic for building kernel is slightly different than other packages. A non-tmpfs
filesystem is overlaid as dpkg-src in this case and build on it rather than tmpfs.

Kernel config is generated in two or three steps during the build process.
make defconfig is run to generate a default config based on architecture, then
the TrueNAS-specific configuration is merged with it, and if a debug kernel is
requested (by uncommenting relevent line in build.sh), additional debugging parameters
are applied.

A customized copy of current kernel source is also mounted as /kernel inside the
package build chroot. A new key has also been added to the build_manifest "kernel_module",
which if set to true will cause the kernel image and headers to be installed in the
package building chroot prior to fulfilling the packages dependencies.

A minor change to behavior in low ram situations (i.e. less than recommended amount) was
added to the effect that update files will not be generated on tmpfs. In testing this
stage of the overall build process was obvserved to be most likely to fail due to running
out of space on the tmpfs filesystem.

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -51,6 +51,9 @@
 	},
 	"base-packages": [
 		"dosfstools",
+		"linux-truenas-libc-dev",
+		"linux-headers-truenas-amd64",
+		"linux-image-truenas-amd64",
 		"consul",
 		"firmware-bnx2",
 		"firmware-bnx2x",
@@ -114,6 +117,8 @@
 	],
 	"iso-packages": [
 		"curl",
+		"bzip2",
+		"linux-image-truenas-amd64",
 		"dialog",
 		"iproute2",
 		"jq",
@@ -127,6 +132,11 @@
 	],
 	"sources": [
 		{
+			"name": "kernel",
+			"repo": "https://github.com/truenas/linux",
+			"branch": "SCALE-v5.10-stable"
+		},
+		{
 			"name": "nfs4xdr_acl_tools",
 			"repo": "https://github.com/truenas/nfs4xdr-acl-tools",
 			"branch": "master"
@@ -134,8 +144,9 @@
 		{
 			"name": "openzfs",
 			"repo": "https://github.com/truenas/zfs",
-			"branch": "truenas/zfs-2.0-release",
+			"branch": "build-with-custom-kernel",
 			"predepscmd": "cp -r contrib/truenas debian",
+			"kernel_module": true,
 			"generate_version": false
 		},
 		{
@@ -180,7 +191,8 @@
 			"name": "scst",
 			"repo": "https://github.com/truenas/scst",
 			"prebuildcmd": "make dpkg DEBEMAIL=no-reply@ixsystems.com DEBFULLNAME=TrueNAS",
-			"branch": "v3.5-branch"
+			"kernel_module": true,
+			"branch": "build-with-truenas-kernel"
 		},
 		{
 			"name": "truenas_binaries",
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -144,7 +144,7 @@
 		{
 			"name": "openzfs",
 			"repo": "https://github.com/truenas/zfs",
-			"branch": "build-with-custom-kernel",
+			"branch": "truenas/zfs-2.0-release",
 			"predepscmd": "cp -r contrib/truenas debian",
 			"kernel_module": true,
 			"generate_version": false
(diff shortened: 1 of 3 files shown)


anodos325:
This is flagged as WIP pending review. So repos are pointed to non-standard branches so that build can be tested.

conf/build.manifest
```diff
@@ -164,7 +175,7 @@
 		{
 			"name": "truenas_installer",
 			"repo": "https://github.com/truenas/truenas-installer",
-			"branch": "master"
+			"branch": "custom-kernel-build"
```
kmoore134, line 178: Will these be changed to master later? If you only want them to test with the pull-request, you can do so just by ensuring that the branch name for this PR is the same as the branch name on those repos. Jenkins / builder will DTRT in that case, and you won't need to modify this branch manually after you merge them all back to respective master.

scripts/build.sh
```diff
@@ -390,41 +439,104 @@ clean_previous_packages() {
...
-		chroot ${DPKG_OVERLAY} apt update || exit_err "Failed apt update"
+mk_kernoverlay() {
+	# Generate kernel overlay (but not mount).
+	# This makes our debian directory and kernel config used for building
+	# debian folder is required to install pre-build dependencies.
+	mkdir ${KERNTMP}
+	mkdir ${KERNWRK}
+	stored_cwd=$(pwd)
+	cp -r ${SOURCES}/kernel/* ${KERNTMP} || exit_err "Failed to copy sources"
+	apt install -y flex bison dwarves libssl-dev > /dev/null
```
kmoore134, line 450: Is it possible that we'll be building other kernel modules later that need different packages installed? Perhaps this should be in the JSON manifest instead for each repo? Like "kernel_pkg_depends": or something?

william-gr (approved):
Other than Kris concerns, lgtm

scripts/build.sh
kmoore134: I'm fine going without this, we can add that functionality later when needed.

kmoore134 (approved):

scripts/build.sh
```diff
@@ -98,7 +128,9 @@ make_bootstrapdir() {
 
 	# Setup our ramdisk, up to 4G should suffice
```
freqlabs, line 129: Could you fix this comment as well when you adjust the branch names please.

scripts/build.sh
```diff
@@ -390,41 +439,104 @@ clean_previous_packages() {
...
+	fi
+	mount -t overlay -o lowerdir="${kernlower}",upperdir="${KERNTMP}",workdir="${KERNWRK}", none "${kernlower}"
+}
+
+umount_kern() {
+	kdir="$1"
+	if [ -z $kdir ]; then
+		kdir="kernel"
+	fi
+        #umount -f "${DPG_OVERLAY}/${kdir}" 2>/dev/null
```
freqlabs, line 488: Use it or lose it ;)

scripts/build.sh
```diff
@@ -445,24 +557,104 @@ build_dpkg() {
...
+	if [ -z "$subarg" -o "$subarg" = "null" ] ; then
+		subdir=""
+	else
+		subdir="/$subarg"
+	fi
+	srcdir="/dpkg-src$subdir"
+	pkgdir="$srcdir/../"
+	do_prebuild "$name" "$predep" "$prebuild" "$subarg" "$generate_version" "$srcdir" "$pkgdir" "false"
+	# Build the package
+	#chroot ${DPKG_OVERLAY} /bin/bash -c "cd $srcdir && debuild $deflags" || exit_err "Failed to build package"
```
freqlabs, line 590: If these commented out commands are worth keeping, please leave a comment explaining why.

scripts/build.sh
```diff
@@ -445,24 +557,104 @@ build_dpkg() {
...
+	else
+		subdir="/$subarg"
+	fi
+	srcdir="/dpkg-src$subdir"
+	pkgdir="$srcdir/../"
 
+	do_prebuild "$name" "$predep" "$prebuild" "$subarg" "$generate_version" "$srcdir" "$pkgdir" "$kmod"
 	# Build the package
 	#chroot ${DPKG_OVERLAY} /bin/bash -c "cd $srcdir && debuild $deflags" || exit_err "Failed to build package"
+	#chroot ${DPKG_OVERLAY} /bin/bash -c "cd $srcdir && debuild $deflags"
```
freqlabs, line 649: This one can definitely go.

freqlabs (approved):

ericbsd (approved):
LGTM

PR #123, opened 2021-03-17
NAS-109487 / 21.04 / Disable nslcd service

themylogin:


themylogin:
retest this please

scripts/build.sh
```diff
@@ -975,6 +975,12 @@ custom_rootfs_setup() {
 		echo '[Install]' >> $file
 		echo 'WantedBy=multi-user.target' >> $file
 	done
+	for file in ${CHROOT_BASEDIR}/tmp/systemd/multi-user.target.wants/*.service; do
```
william-gr, line 978: So you want to exclude every file but rrdcached.service? Curious why?

@kmoore134 you being the shell expert can think of a smarter way to do this? Maybe find with `-name \! "rrdcached.service -delete`?

themylogin: If a unit is in `multi-user.target.wants`, there is no way to disable it (you run `service disable nslcd`, but it is still started after a reboot). With `rrdcached` it is not the case, we don't want to disable it.

william-gr (approved):

PR #124, opened 2021-03-23
Work around broken alien package

freqlabs:
OpenZFS uses alien to build the debs, but Debian has a bug in their
alien version 8.95.3 that causes the generated rules to be incorrect.

Work around this for the time being with a local version of the alien
package.  This has been named alienz to make sure we are using the
local version rather than the upstream package.

https://salsa.debian.org/debian/alien/-/merge_requests/2

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -146,6 +146,9 @@ sources:
 - name: nfs4xdr_acl_tools
   repo: https://github.com/truenas/nfs4xdr-acl-tools
   branch: master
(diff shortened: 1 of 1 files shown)


freqlabs:
This works in conjunction with a change I have pushed to truenas/zfs that blacklists the broken alien version by changing the build dependency to `alien (<< 8.95.3) | alien (>> 8.95.3) | alienz`

kmoore134 (approved):

PR #125, opened 2021-03-25
Do not use overlay mount

sonicaj:
This commit adds changes to not use overlay mount for building kernel package as this means that this won't build on zfs ( https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1873917 ) because we are unable to mount.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -40,7 +40,6 @@ HAS_LOW_RAM=0
 # Debug kernel can be built by uncommenting DEBUG_KERNEL
 
 KERNTMP="./tmp/kern"
-KERNWRK="./tmp/kernwrk"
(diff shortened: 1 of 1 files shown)


william-gr:
Lets ask @anodos325 but I don't think thats the best course of action.

Perhaps make it behave differently for ZFS only, for now.

sonicaj:
@william-gr I consulted with @anodos325 and he agrees we can go with this approach. I was also initially thinking that we could maybe use bind mount when on zfs. @anodos325 can you please confirm ?

anodos325 (approved):
Works in my builder, so I'm happy if everyone else is.

william-gr (approved):

PR #129, opened 2021-03-30
Add Prometheus node_exporter

SuperQ:
* Add prometheus-node-exporter package.
* Configure to listen on localhost by default.
* Disable rapl collector due to noisy log lines.
* Add an rsync to copy over /etc/ file overrides.

https://jira.ixsystems.com/browse/NAS-103563

Signed-off-by: Ben Kochie <superq@gmail.com>

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -115,6 +115,8 @@ additional-packages:
   comment: requested by sales (NAS-108787)
 - package: dnsutils
   comment: requested by community (NAS-109391)
+- package: prometheus-node-exporter
+  comment: requested by community (NAS-103563)
 
 #
 # List of additional packages installed into TrueNAS SCALE ISO file
--- a/conf/etc/default/prometheus-nod-exporter
+++ b/conf/etc/default/prometheus-nod-exporter
@@ -1,0 +1,6 @@
+# Configure node_exporter default flags
+#
+# * Listen on localhost by default.
+# * Disable rapl collector due to noisy log lines.
+ARGS="--web.listen-address=[::1]:9100 --no-collector.rapl"
+
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -984,6 +984,9 @@ custom_rootfs_setup() {
 			|| exit_err "Failed wget of nomad"
 	fi
 	unzip -d ${CHROOT_BASEDIR}/usr/bin ${CACHE_DIR}/nomad_${NOMADVER}.zip || exit_err "Failed unzip of nomad"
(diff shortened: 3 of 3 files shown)


conf/build.manifest
```diff
@@ -72,6 +72,7 @@ base-packages:
 - truenas-samba
 - nfs4xdr-acl-tools
 - netatalk
+- prometheus-node-exporter
```
Ornias1993, line 75: This needs to go under "additional-packages" instead....
And preferably make a Jira suggestion first and link that one to both the PR and the package...

Like this:

```
- package: prometheus-node-exporter
  comment: requested by community (NAS-number)
```

Ornias1993 (changes requested):
Looks pretty kiss, but the package is added in the wrong place... (as it isn't a base-feature).

Also would be nice to make the Jira ticket first, link the PR (by editing the PR title to follow the iX PR format) and linking the package to the jira ticket like explained below...

conf/build.manifest
SuperQ: Sounds good, I found `NAS-103563`. Perhaps we can re-open it.

Ornias1993 (approved):
Looks good to me...

Just a slight note:
Change the title to:
`NAS-103563 / 21.06 / Add Prometheus node_exporter`

(it does some fancy detection of the NAS-* number to connect it to the right Jira ticket ;-) )

SuperQ:
So, after some thought and discussion, maybe we should skip this in favor of doing it via Helm containers.

Since there's some desire to avoid the Prometheus CRD, there is a dedicated node_exporter chart that could be used:

https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus-node-exporter

lapo-luchini:
BTW TrueNAS can already be monitored on Grafana using PromQL when using a VictoriaMetrics datasource (which accepts Graphite data as well as Prometheus data) by using the official Graphite export.
[I'm creating a dashboard for that](https://github.com/lapo-luchini/grafana-truenas).

Ornias1993:
> BTW TrueNAS can already be monitored on Grafana using PromQL when using a VictoriaMetrics datasource (which accepts Graphite data as well as Prometheus data) by using the official Graphite export.
> [I'm creating a dashboard for that](https://github.com/lapo-luchini/grafana-truenas).

This isn't about monitoring TrueNAS itself, well not in such a way that it would part of the current solutions. This has to do with SCALE K8S metrics. Not with TrueNAS itself.

PR #126, opened 2021-03-25
NAS-107508 / 21.04 / Use grub2 from custom repo

themylogin:

PR #133, opened 2021-03-31
Revert "NAS-107508 / 21.04 / Use grub2 from custom repo"

william-gr:
Reverts truenas/scale-build#126


william-gr:
@themylogin breaking the build. Reverted for now

PR #139, opened 2021-04-20
NAS-108217 / 21.06 / Refactor builder in python

sonicaj:


scale_build/bootstrap/cache.py
```diff
@@ -0,0 +1,73 @@
+import os
+
+from scale_build.clean import clean_packages
+from scale_build.exceptions import CallError
+
+from .hash import get_all_repo_hash, get_base_hash
+from .logger import get_logger
```
themylogin, line 7: Why do you need so many log files and three functions named `get_logger`?

scale_build/bootstrap/cache.py
```diff
@@ -0,0 +1,73 @@
...
+    logger = get_logger(cache_type)
+    logger.debug('Removing base chroot cache for %s', cache_type)
+    for path in map(
+        lambda p: os.path.join(CACHE_DIR, p), (get_cache_filename(cache_type), get_cache_hash_filename(cache_type))
+    ):
+        if os.path.exists(path):
+            os.unlink(path)
+
+
+def restore_basecache(cache_type, chroot_basedir, logger=None):
```
themylogin, line 49: What is this `logger=None` for? I don't see any callers that omit this argument

scale_build/bootstrap/cache.py
```diff
@@ -0,0 +1,73 @@
...
+    run([
+        'unsquashfs', '-f', '-d', chroot_basedir, os.path.join(CACHE_DIR, get_cache_filename(cache_type))
+    ], exception=CallError, exception_msg='Failed unsquashfs', logger=logger)
```
themylogin, line 52: What is the pleasure of writing `logger=logger` every time you want to run a command?

scale_build/bootstrap/configure.py
```diff
@@ -0,0 +1,101 @@
...
+from .utils import BUILDER_DIR, CACHE_DIR, CHROOT_BASEDIR, get_apt_preferences, get_manifest, run
+
+
+def make_bootstrapdir(bootstrapdir_type, logger_file=None):
+    assert bootstrapdir_type in ('cdrom', 'package')
+    remove_boostrap_directory()
+    try:
+        _make_bootstrapdir_impl(bootstrapdir_type, logger_file)
+    finally:
+        remove_boostrap_directory()
```
themylogin, line 18: Doesn't this function really make a cached bootstrap image? It seems weird that `make_something` calls `remove_something` at the very end? Or are `bootstrapdir` and `boostrap_directory` different things?

scale_build/bootstrap/configure.py
```diff
@@ -0,0 +1,101 @@
...
+
+    # Check if we should invalidate the base cache
+    if validate_basecache(bootstrapdir_type):
+        logger.debug('Basechroot cache is intact and does not need to be changed')
+        return
+
+    run([
+        'apt-key', '--keyring', '/etc/apt/trusted.gpg.d/debian-archive-truenas-automatic.gpg', 'add',
+        os.path.join(BUILDER_DIR, 'keys/truenas.gpg')
+    ], exception=CallError, exception_msg='Failed adding truenas.gpg apt-key', **run_args)
```
themylogin, line 37: Do we have any `exception` other than `CallError`?

scale_build/bootstrap/configure.py
```diff
@@ -0,0 +1,101 @@
...
+        ['debootstrap'] + deopts + [
+            '--keyring', '/etc/apt/trusted.gpg.d/debian-archive-truenas-automatic.gpg', 'bullseye',
+            CHROOT_BASEDIR, apt_repos['url']
+        ], exception=CallError, exception_msg='Failed debootstrap', **run_args
+    )
+
+    create_basehash(bootstrapdir_type)
+    os.makedirs(os.path.join(CACHE_DIR, 'apt'), exist_ok=True)
+
+    run(['mount', 'proc', os.path.join(CHROOT_BASEDIR, 'proc'), '-t', 'proc'], **run_args)
```
themylogin, line 50: Do we need all this `os.path.join`? `f'{CHROOT_BASEDIR}/proc'` seems more compact

scale_build/bootstrap/hash.py
```diff
@@ -0,0 +1,40 @@
...
+
+from scale_build.exceptions import CallError
+
+from .utils import CHROOT_BASEDIR, get_apt_preferences, get_manifest, run
+
+
+to_disable = ('requests', 'urllib3')
+for name in filter(lambda k: k.startswith(to_disable), logging.Logger.manager.loggerDict.keys()):
+    custom_logger = logging.getLogger(name)
+    custom_logger.disabled = True
```
themylogin, line 15: Why disable a logger instead of just setting an appropriate log level? Why are you walking all loggers instead of just using propagation?

scale_build/bootstrap/hash.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def get_repo_hash(repo_url, distribution):
+    resp = requests.get(urllib.parse.urljoin(repo_url, os.path.join('dists', distribution, 'Release')))
```
themylogin, line 19: `timeout`?

scale_build/bootstrap/utils.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def get_apt_preferences():
+    return '\n\n'.join(
+        '\n'.join(f'{k}: {v}' for k, v in pref.items()) for pref in get_manifest()['apt_preferences']
+    )
+
+
+def get_cache_filename(cache_type):
+    return f'basechroot-{cache_type}.squashfs'
```
themylogin, line 13: Are `basechroot` and `cache` another names for something we already have above in this diff?

scale_build/config.py
```diff
@@ -0,0 +1,21 @@
+import os
+import time
+
+from datetime import datetime
+
+
+BUILD_TIME = int(time.time())
+BUILD_TIME_OBJ = datetime.fromtimestamp(BUILD_TIME)
+BUILDER_DIR = os.getenv('BUILDER_DIR', './')
+BRANCH_OVERRIDES = {k[:-(len('_OVERRIDE'))]: v for k, v in os.environ.items() if k.endswith('_OVERRIDE')}
```
themylogin, line 10: Should we fail if an unknown package is specified here?

scale_build/checkout.py
```diff
@@ -0,0 +1,32 @@
...
+        gh_override = BRANCH_OVERRIDES.get(package.name)
+
+        # TRY_BRANCH_OVERRIDE is a special use-case. It allows setting a branch name to be used
+        # during the checkout phase, only if it exists on the remote.
+        #
+        # This is useful for PR builds and testing where you want to use defaults for most repos
+        # but need to test building of a series of repos with the same experimental branch
+        #
+        if TRY_BRANCH_OVERRIDE:
+            cp = run(['git', 'ls-remote', package.origin], check=False)
```
themylogin, line 28: So if there is a one-time network error it will silently ignore the existence of `TRY_BRANCH_OVERRIDE` on the remote and use default branch instead?

scale_build/config.py
```diff
@@ -0,0 +1,21 @@
...
+CODE_NAME = os.getenv('CODENAME', 'Angelfish')
```
themylogin, line 11: Shouldn't things like this (`Angelfish`) be in the manifest?

scale_build/config.py
```diff
@@ -0,0 +1,21 @@
...
+PARALLEL_BUILD = int(os.getenv('PARALLEL_BUILDS', max(os.cpu_count(), 8) / 4))
```
themylogin, line 12: Is this to ensure we always have parallel builds even on systems with 4 CPUs?

scale_build/exceptions.py
```diff
@@ -0,0 +1,26 @@
+import errno
+
+
+class CallError(Exception):
+    def __init__(self, errmsg, errno=errno.EFAULT, extra=None):
```
themylogin, line 5: Do we really need errno and extra?

scale_build/image/bootstrap.py
```diff
@@ -0,0 +1,43 @@
...
+import shutil
+
+from scale_build.bootstrap.cache import restore_basecache
+from scale_build.utils.run import run
+from scale_build.utils.paths import CD_DIR, CHROOT_BASEDIR, PKG_DIR, RELEASE_DIR, TMPFS
+
+from .utils import PACKAGE_PATH
+
+
+def setup_chroot_basedir(basecache_type, logger=None):
```
themylogin, line 11: Can we make all these `setup` and `cleanup` functions context managers?

scale_build/image/iso.py
```diff
@@ -0,0 +1,103 @@
...
+
+    # Set default PW to root
+    run_in_chroot(r'echo -e \"root\nroot\" | passwd root', logger=iso_logger)
+
+    # Create /etc/version
+    with open(os.path.join(CHROOT_BASEDIR, 'etc/version'), 'w') as f:
+        f.write(VERSION)
+
+    # Copy the CD files
+    for source, destination in (
```
themylogin, line 43: Can't we just copy all `cd-files` we have without listing them here?

scale_build/image/iso.py
```diff
@@ -0,0 +1,103 @@
...
+    os.makedirs(os.path.join(CD_DIR, 'live'), exist_ok=True)
+    shutil.move(tmp_truenas_path, os.path.join(CD_DIR, 'live/filesystem.squashfs'))
+
+    # Copy over boot and kernel before rolling CD
+    shutil.copytree(os.path.join(CHROOT_BASEDIR, 'boot'), os.path.join(CD_DIR, 'boot'))
+    # Dereference /initrd.img and /vmlinuz so this ISO can be re-written to a FAT32 USB stick using Windows tools
+    shutil.copy(os.path.join(CHROOT_BASEDIR, 'initrd.img'), CD_DIR)
+    shutil.copy(os.path.join(CHROOT_BASEDIR, 'vmlinuz'), CD_DIR)
+    for f in itertools.chain(
+        glob.glob(os.path.join(CD_DIR, 'boot/initrd.img-')),
```
themylogin, line 72: `boot/initrd.img-*`

scale_build/image/update.py
```diff
@@ -0,0 +1,158 @@
...
+    # Generate native systemd unit files for SysV services that lack ones to prevent systemd-sysv-generator warnings
+    tmp_systemd = os.path.join(CHROOT_BASEDIR, 'tmp/systemd')
+    os.makedirs(tmp_systemd)
+    run_in_chroot(
+        '/usr/lib/systemd/system-generators/systemd-sysv-generator /tmp/systemd /tmp/systemd /tmp/systemd',
+        rootfs_logger
+    )
+    for unit_file in filter(lambda f: f.endswith('.service'), os.listdir(tmp_systemd)):
+        with open(os.path.join(tmp_systemd, unit_file), 'a') as f:
+            f.write(textwrap.dedent('''
```
themylogin, line 115: ```'''\```

scale_build/image/update.py
```diff
@@ -0,0 +1,158 @@
...
+                [Install]
+                WantedBy=multi-user.target
+            '''))
+
+    for file_path in map(
```
themylogin, line 120: Bash code that was doing the same thing was way nicer. This feels wrong.

scale_build/image/update.py
```diff
@@ -0,0 +1,158 @@
...
+    run_in_chroot('apt autoremove -y', rootfs_logger, 'Failed atp autoremove')
+
+    # We install the nvidia-kernel-dkms package which causes a modprobe file to be written
+    # (i.e /etc/modprobe.d/nvidia.conf). This file tries to modprobe all the associated
+    # nvidia drivers at boot whether or not your system has an nvidia card installed.
+    # For all truenas certified and truenas enterprise hardware, we do not include nvidia GPUS.
+    # So to prevent a bunch of systemd "Failed" messages to be barfed to the console during boot,
+    # we remove this file because the linux kernel dynamically loads the modules based on whether
+    # or not you have the actual hardware installed in the system.
+    if os.path.exists(os.path.join(CHROOT_BASEDIR, 'etc/modprobe.d/nvidia.conf')):
```
themylogin, line 149: If you use `with contextlib.suppress(FileNotFoundError)`, you won't have to repeat the same name twice

scale_build/image/update.py
```diff
@@ -0,0 +1,158 @@
...
+        os.unlink(os.path.join(CHROOT_BASEDIR, 'etc/modprobe.d/nvidia.conf'))
+
+    for path in (
+        os.path.join(CHROOT_BASEDIR, 'usr/share/doc'),
+        os.path.join(CHROOT_BASEDIR, 'var/cache/apt'),
+        os.path.join(CHROOT_BASEDIR, 'var/lib/apt/lists'),
+    ):
+        shutil.rmtree(path, ignore_errors=True)
```
themylogin, line 157: Which errors are you ignoring?

scale_build/image/utils.py
```diff
@@ -0,0 +1,16 @@
...
+from scale_build.utils.run import run
+from scale_build.utils.paths import CHROOT_BASEDIR
+
+
+PACKAGE_PATH = os.path.join(CHROOT_BASEDIR, 'packages')
+
+
+def run_in_chroot(command, logger=None, exception_message=None, **kwargs):
+    return run(
+        f'chroot {CHROOT_BASEDIR} /bin/bash -c "{command}"', shell=True, logger=logger,
```
themylogin, line 14: What if `command` contains `"`?

themylogin (changes requested):

scale_build/image/iso.py
sonicaj: They have specific source / destination, please advise if you think we should be perhaps re-arranging cd folder so that we can copy it directly in the chroot ?

scale_build/config.py
sonicaj: Yes this was to ensure that with machine less then 8 cpus, we have parallel builds by default unless overridden

scale_build/config.py
sonicaj: I don't think so, we won't be consuming it.

scale_build/bootstrap/hash.py
sonicaj: I have tried disabling propagation but it did not help. And our default log level is debug, so these logged messaged are still caught by us.

scale_build/preflight.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+WANTED_PACKAGES = {
+    'make',
+    'debootstrap',
+    'git',
+    'xorriso',
+    'grub-mkrescue',
```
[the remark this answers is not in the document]
themylogin: Can you remove xorriso and grub-mkrescue please? We should not be running them on host. And we are not (I changed this a couple of month ago) doing so.

scale_build/bootstrap/hash.py
themylogin: `logging.getLogger("requests").setLevel(logging.INFO)` should do the job

scale_build/config.py
themylogin: I think we should fail if there is an unconsumed environment variable (e.g. due to a typo)

scale_build/config.py
themylogin: Why? Parallel build testing purposes?

scale_build/config.py
sonicaj: Any unconsumed env variable ?

scale_build/image/iso.py
themylogin: Yeah, I think that'll be easier to use in the future

scale_build/config.py
sonicaj: There could be more then just this example

scale_build/config.py
sonicaj: Well from my tests, i see that when machines with 8 cpus can still see performance gains with 2 threads whereas when we increase this number, we should be dividing by 4 as some packages are already consuming all cores like kernel/openzfs.

scale_build/config.py
themylogin: On all that get into `BRANCH_OVERRIDES`. E.g. `MILDLEWARE_OVERRIDE=xxx`

scale_build/bootstrap/hash.py
sonicaj: It does not
```
[2021-04-23 04:00:35] Starting new HTTP connection (1): apt.tn.ixsystems.com:80
[2021-04-23 04:00:35] http://apt.tn.ixsystems.com:80 "GET /apt-direct/unstable/debian/dists/bullseye/Release HTTP/1.1" 200 7522
[2021-04-23 04:00:35] Starting new HTTP connection (1): apt.tn.ixsystems.com:80
[2021-04-23 04:00:36] http://apt.tn.ixsystems.com:80 "GET /apt-direct/unstable/debian-sid/dists/sid/Release HTTP/1.1" 200 7511
[2021-04-23 04:00:36] Starting new HTTP connection (1): apt.tn.ixsystems.com:80
[2021-04-23 04:00:36] http://apt.tn.ixsystems.com:80 "GET /apt-direct/unstable/docker/dists/buster/Release HTTP/1.1" 200 3486
[2021-04-23 04:00:36] Starting new HTTP connection (1): apt.tn.ixsystems.com:80
[2021-04-23 04:00:37] http://apt.tn.ixsystems.com:80 "GET /apt-direct/unstable/gluster/dists/bullseye/Release HTTP/1.1" 200 3446
[2021-04-23 04:00:37] Starting new HTTP connection (1): apt.tn.ixsystems.com:80
[2021-04-23 04:00:37] http://apt.tn.ixsystems.com:80 "GET /apt-direct/unstable/libnvidia/dists/bullseye/Release HTTP/1.1" 200 3483
[2021-04-23 04:00:37] Starting new HTTP connection (1): apt.tn.ixsystems.com:80
[2021-04-23 04:00:38] http://apt.tn.ixsystems.com:80 "GET /apt-direct/unstable/nvidia-container/dists/bullseye/Release HTTP/1.1" 200 3495
[2021-04-23 04:00:38] Starting new HTTP connection (1): apt.tn.ixsystems.com:80
[2021-04-23 04:00:38] http://apt.tn.ixsystems.com:80 "GET /apt-direct/unstable/nvidia-docker/dists/bullseye/Release HTTP/1.1" 200 3481
[2021-04-23 04:00:38] Starting new HTTP connection (1): apt.tn.ixsystems.com:80
```

themylogin: It does
```
import logging
import requests
logging.basicConfig(level=logging.DEBUG)
logging.getLogger("urllib3").setLevel(logging.INFO)
requests.get("http://google.com")
```
Prints nothing

Makefile
```diff
@@ -1,16 +1,28 @@
...
+PYTHON?=/usr/bin/python3
+COMMIT_HASH=$(shell git rev-parse --short HEAD)
+
+check:
+ifeq ("$(wildcard ./venv-${COMMIT_HASH})","")
+	@rm -rf venv-*
+	@${PYTHON} -m pip install -U virtualenv >/dev/null 2>&1
+	@${PYTHON} -m venv venv-${COMMIT_HASH}
+	@. ./venv-${COMMIT_HASH}/bin/activate && \
+		python3 -m pip install -r requirements.txt >/dev/null 2>&1 && \
```
themylogin, line 13: Will this fail silently in case of an error?

scale_build/bootstrap/bootstrapdir.py
```diff
@@ -0,0 +1,149 @@
...
+        if self.mirror_cache_intact:
+            # Mirror cache is intact, we do not need to re-create the bootstrap directory
+            self.logger.debug('Basechroot cache is intact and does not need to be changed')
+            return
+
+        self.add_trusted_apt_key()
+        apt_repos = get_manifest()['apt-repos']
+        self.run(
+            ['debootstrap'] + self.deopts + [
+                '--keyring', '/etc/apt/trusted.gpg.d/debian-archive-truenas-automatic.gpg', 'bullseye',
```
themylogin, line 30: Shouldn't `bullseye` be somewhere in the manifest instead of being hardcoded?

scale_build/bootstrap/bootstrapdir.py
```diff
@@ -0,0 +1,149 @@
...
+    def clean_mounts(self):
+        for command in (
+            ['umount', '-f', os.path.join(self.chroot_basedir, 'proc')],
+            ['umount', '-f', os.path.join(self.chroot_basedir, 'sys')],
+        ):
+            run(command, check=False)
+
+    def clean_setup(self):
+        self.clean_mounts()
+        shutil.rmtree(self.chroot_basedir, ignore_errors=True)
```
themylogin, line 107: What errors are you ignoring?

We expect `self.chroot_basedir` to be non-existent after this method terminates successfully. Will it always be so?

scale_build/bootstrap/cache.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @property
+    def mirror_cache_intact(self):
+        intact = True
+        if not self.cache_exists:
+            # No hash file? Lets remove to be safe
+            intact = False
+            self.logger.debug('Cache does not exist')
+
+        if get_all_repo_hash() != self.get_mirror_cache():
```
themylogin, line 50: Shouldn't it be `elif` here?

scale_build/bootstrap/hash.py
```diff
@@ -0,0 +1,79 @@
...
+
+logging.getLogger('urllib3').setLevel(logging.INFO)
+
+
+INSTALLED_PACKAGES_REGEX = re.compile(r'([^\t]+)\t([^\t]+)\t([\S]+)\n')
+
+
+def get_repo_hash(repo_url, distribution):
+    resp = requests.get(urllib.parse.urljoin(repo_url, os.path.join('dists', distribution, 'Release')), timeout=60)
+    if resp.status_code != 200:
```
themylogin, line 25: `resp.raise_for_status()`

scale_build/checkout.py
```diff
@@ -0,0 +1,43 @@
...
+                    if branch_exists_in_repository(package.origin, TRY_BRANCH_OVERRIDE):
+                        gh_override = TRY_BRANCH_OVERRIDE
+                except CallError:
+                    retries -= 1
+                    logger.debug(
+                        'Failed to determine if %r branch exists for %r. Trying again',
+                        TRY_BRANCH_OVERRIDE, package.origin
+                    )
+                    if not retries:
+                        logger.debug('Unable to determine if %r branch exists in 3 attempts.')
```
themylogin, line 39: `%r` is not used

Why can't we just fail here? If I specified `TRY_BRANCH_OVERRIDE`, I want my build to use `TRY_BRANCH_OVERRIDE` or to fail if it does not exist. I don't want it to fake a build by giving me an image that uses some other branch. Will I have to read logs every time to see if it gave me a build I ordered?

scale_build/clean.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def clean_bootstrap_logs():
+    for f in filter(lambda f: f.startswith('bootstrap'), os.listdir(LOG_DIR)):
+        os.unlink(os.path.join(LOG_DIR, f))
+
+
+def clean_packages():
+    for d in (HASH_DIR, PKG_DIR):
+        shutil.rmtree(d, ignore_errors=True)
```
themylogin, line 18: What errors are you ignoring?

We expect `HASH_DIR` and `PKG_DIR` to be non-existent after this method terminates successfully. Will it always be so?

scale_build/clean.py
```diff
@@ -0,0 +1,26 @@
...
+        os.makedirs(d)
+
+
+def complete_cleanup():
+    for path in (LOG_DIR, SOURCES_DIR, TMP_DIR):
+        shutil.rmtree(path, ignore_errors=True)
```
themylogin, line 24: What errors are you ignoring?

We expect `LOG_DIR`, `SOURCES_DIR` and `TMP_DIR` to be non-existent after this method terminates successfully. Will it always be so?

scale_build/image/bootstrap.py
```diff
@@ -0,0 +1,42 @@
...
+import shutil
+
+from scale_build.utils.run import run
+from scale_build.utils.paths import CD_DIR, CHROOT_BASEDIR, PKG_DIR, RELEASE_DIR, TMPFS
+
+from .utils import PACKAGE_PATH
+
+
+def setup_chroot_basedir(bootstrapdir_obj, logger):
+    shutil.rmtree(CHROOT_BASEDIR, ignore_errors=True)
```
themylogin, line 11: What errors are you ignoring?

We expect `CHROOT_BASEDIR` to be non-existent after this method terminates successfully. Will it always be so?

scale_build/image/bootstrap.py
```diff
@@ -0,0 +1,42 @@
...
+    )
+    bootstrapdir_obj.restore_cache(CHROOT_BASEDIR)
+    run(['mount', 'proc', os.path.join(CHROOT_BASEDIR, 'proc'), '-t', 'proc'], logger=logger)
+    run(['mount', 'sysfs', os.path.join(CHROOT_BASEDIR, 'sys'), '-t', 'sysfs'], logger=logger)
+    os.makedirs(PACKAGE_PATH, exist_ok=True)
+    run(['mount', '--bind', PKG_DIR, PACKAGE_PATH], logger=logger)
+
+
+def umount_tmpfs_and_clean_chroot_dir():
+    shutil.rmtree(CHROOT_BASEDIR, ignore_errors=True)
```
themylogin, line 25: What errors are you ignoring?

We expect `CHROOT_BASEDIR` to be non-existent after this method terminates successfully. Will it always be so?

scale_build/image/iso.py
```diff
@@ -0,0 +1,94 @@
...
+    # Create /etc/version
+    with open(os.path.join(CHROOT_BASEDIR, 'etc/version'), 'w') as f:
+        f.write(VERSION)
+
+    # Copy the CD files
+    distutils.dir_util._path_created = {}
+    distutils.dir_util.copy_tree(CD_FILES_DIR, CHROOT_BASEDIR, preserve_symlinks=True)
+
+    # Create the CD assembly dir
+    shutil.rmtree(CD_DIR, ignore_errors=True)
```
themylogin, line 43: What errors are you ignoring?

We expect `CD` to be at least empty after this method terminates successfully. Will it always be so?

scale_build/image/iso.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def prune_cd_basedir():
+    for path in itertools.chain([
+        os.path.join(CHROOT_BASEDIR, 'var/cache/apt'),
+        os.path.join(CHROOT_BASEDIR, 'var/lib/apt'),
+        os.path.join(CHROOT_BASEDIR, 'usr/share/doc'),
+        os.path.join(CHROOT_BASEDIR, 'usr/share/man'),
+    ] + glob.glob(os.path.join(CHROOT_BASEDIR, 'lib/modules/*-amd64/kernel/sound'))):
+        shutil.rmtree(path, ignore_errors=True)
```
themylogin, line 94: What errors are you ignoring?

We expect all of these directories to be at least empty after this method terminates successfully. Will it always be so?

scale_build/image/update.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+def build_rootfs_image(update_image_logger):
+    for f in glob.glob(os.path.join('./tmp/release', '*.update*')):
+        os.unlink(f)
+
+    shutil.rmtree(UPDATE_DIR, ignore_errors=True)
```
themylogin, line 25: What errors are you ignoring?

We expect `UPDATE_DIR` to be at least empty after this method terminates successfully. Will it always be so?

scale_build/package.py
```diff
@@ -0,0 +1,167 @@
...
+    package_logger = get_logger('build_packages', 'build_packages.log', 'w')
+    logger.info('Building packages (%s/build_packages.log)', LOG_DIR)
+    logger.debug('Setting up bootstrap directory')
+
+    with PackageBootstrapDirectory(package_logger) as p:
+        p.setup()
+
+    logger.debug('Successfully setup bootstrap directory')
+
+    shutil.rmtree(PKG_LOG_DIR, ignore_errors=True)
```
themylogin, line 112: What errors are you ignoring?

We expect `PKG_LOG_DIR` to be non-existent after this method terminates successfully. Will it always be so?

scale_build/packages/overlay.py
```diff
@@ -0,0 +1,77 @@
...
+            ['umount', '-R', '-f', self.dpkg_overlay],
+            ['umount', '-R', '-f', self.tmpfs_path],
+        ):
+            run(command, check=False)
+
+        for path in (
+            self.chroot_overlay, self.dpkg_overlay, self.workdir_overlay, self.chroot_base_directory,
+            self.sources_overlay, self.tmpfs_path
+        ):
+            shutil.rmtree(path, ignore_errors=True)
```
themylogin, line 77: What errors are you ignoring?

We expect all these directories to be at least empty after this method terminates successfully. Will it always be so?

scale_build/packages/package.py
```diff
@@ -0,0 +1,189 @@
...
+        branch = branch_override or self.branch
+        git_logger = get_logger(f'checkout_{self.name}', 'git-checkout.log', 'w')
+        if branch == self.existing_branch and self.origin == origin_url:
+            logger.debug('Updating git repo [%s] (%s)', self.name, GIT_LOG_PATH)
+            run(['git', '-C', self.source_path, 'fetch', '--unshallow'], logger=git_logger, check=False)
+            run(['git', '-C', self.source_path, 'fetch', 'origin', branch], logger=git_logger)
+            run(['git', '-C', self.source_path, 'reset', '--hard', f'origin/{branch}'], logger=git_logger)
+        else:
+            logger.debug('Checking out git repo [%s] (%s)', self.name, GIT_LOG_PATH)
+            shutil.rmtree(self.source_path, ignore_errors=True)
```
themylogin, line 176: What errors are you ignoring?

We expect `self.source_path` to be non-existent after this method terminates successfully. Will it always be so?

scale_build/exceptions.py
```diff
@@ -0,0 +1,24 @@
+import errno
+
+
+class CallError(Exception):
+    def __init__(self, errmsg):
+        self.errmsg = errmsg
+
+    def __str__(self):
+        return f'[{errno.EFAULT}] {self.errmsg}'
```
themylogin, line 9: Do we need to prefix all our error messages with `[14] `?

scale_build/image/iso.py
```diff
@@ -0,0 +1,94 @@
...
+
+    shutil.copy(UPDATE_FILE, os.path.join(CD_DIR, 'TrueNAS-SCALE.update'))
+    os.makedirs(os.path.join(CHROOT_BASEDIR, RELEASE_DIR), exist_ok=True)
+    os.makedirs(os.path.join(CHROOT_BASEDIR, CD_DIR), exist_ok=True)
+
+    run(['mount', '--bind', RELEASE_DIR, os.path.join(CHROOT_BASEDIR, RELEASE_DIR)])
+    run(['mount', '--bind', CD_DIR, os.path.join(CHROOT_BASEDIR, CD_DIR)])
+
+    run_in_chroot(['apt-get', 'update'], logger=iso_logger, check=False)
+    run_in_chroot(
```
themylogin, line 74: Why `check=False`?

scale_build/image/iso.py
```diff
@@ -0,0 +1,94 @@
...
+        ['apt-get', 'install', '-y', 'grub-efi', 'grub-pc-bin', 'mtools', 'xorriso'], logger=iso_logger, check=False
+    )
+    run_in_chroot(
+        ['grub-mkrescue', '-o', os.path.join(RELEASE_DIR, f'TrueNAS-SCALE-{VERSION}.iso'), CD_DIR], logger=iso_logger
+    )
+    run(['umount', '-f', os.path.join(CHROOT_BASEDIR, CD_DIR)])
```
themylogin, line 80: Will we keep those mounted if `grub-mkrescue` fails?

scale_build/image/update.py
```diff
@@ -0,0 +1,147 @@
...
+    # Do any custom rootfs setup
+    custom_rootfs_setup(update_image_logger)
+
+    # Do any pruning of rootfs
+    clean_rootfs(update_image_logger)
+
+    # Copy the default sources.list file
+    shutil.copy(CONF_SOURCES, os.path.join(CHROOT_BASEDIR, 'etc/apt/sources.list'))
+
+    run_in_chroot(['depmod'], update_image_logger, check=False)
```
themylogin, line 87: Shouldn't it be in `custom_rootfs_setup`?

scale_build/image/update.py
```diff
@@ -0,0 +1,147 @@
...
+        '/usr/lib/systemd/system-generators/systemd-sysv-generator', '/tmp/systemd', '/tmp/systemd', '/tmp/systemd'
+    ], rootfs_logger)
+    for unit_file in filter(lambda f: f.endswith('.service'), os.listdir(tmp_systemd)):
+        with open(os.path.join(tmp_systemd, unit_file), 'a') as f:
+            f.write(textwrap.dedent('''\
+                [Install]
+                WantedBy=multi-user.target
+            '''))
+
+    run(
```
themylogin, line 115: Can't this be done in python?

scale_build/package.py
```diff
@@ -0,0 +1,167 @@
...
+    clean_bootstrap_logs()
+    _build_packages_impl()
+
+
+def _build_packages_impl():
+    package_logger = get_logger('build_packages', 'build_packages.log', 'w')
+    logger.info('Building packages (%s/build_packages.log)', LOG_DIR)
+    logger.debug('Setting up bootstrap directory')
+
+    with PackageBootstrapDirectory(package_logger) as p:
```
themylogin, line 107: What is the point of it being a context manager?

themylogin (changes requested):

scale_build/bootstrap/bootstrapdir.py
sonicaj: I am not sure I follow, why it wouldn't be after this ? We ignore it so that we can safely call it before setting up bootstrap directory to make sure we have a clean start.

themylogin: If `shutil.rmtree` fails to remove something inside `self.chroot_basedir`, it won't be able to remove `self.chroot_basedir` too due to it being non-empty, but as `ignore_errors=True` is specified, this will be ignored and no exception will be raised.

scale_build/checkout.py
sonicaj: My intention was not to change existing behavior. The builder relies on this behavior right now for generating builds. Can you please let me know how you think we should perhaps cater to this use case ? Otherwise the user would have to specify an override for each repo ( we have ~44 almost )

themylogin: > Otherwise the user would have to specify an override for each repo ( we have ~44 almost )

Why?

sonicaj: I am not sure why this was added but i think the use case was to see if any of the sources have `TRY_BRANCH_OVERRIDE` branch and if they do use it so that we have a build which covers the ticket in question. Like say the value set is NAS-11111, now if the ticket has changes to middleware and some other repos, just specifying this in the builder PR, we can expect the resulting build to have all the changes from all these repos. @kmoore134 can you please confirm ?

kmoore134: That is used by CI. SO that if you issue a pull-request to truenas/scale-build for branch NAS-111111, and there is an identical branch on truenas/middleware called NAS-111111, it'll use same branches on both for that build. Thats how to handle complex PR's that rely on other WiP branches or PRs to get test runs.

scale_build/bootstrap/bootstrapdir.py
sonicaj: Understood. My only use case was to be able to call these endpoints even if the directories did not exist but yes, ignoring all errors is not wise. I'll make required changes for all such usages. Thank you for pointing it out.

scale_build/package.py
sonicaj: So that it properly cleans up relevant mount points in case of a failure ?

scale_build/image/update.py
sonicaj: I changed it as you advised the bash code was nicer ?

scale_build/image/iso.py
sonicaj: No, they will be cleaned if `make_iso_file` function fails

scale_build/image/update.py
sonicaj: ( https://github.com/truenas/scale-build/pull/139/commits/13eb1435040eb479fd37a1c967f537dd0168bdf4#diff-bcc5f33a65d37e234ad4bd1e940546df9002f3f5b84efb2b01a99526695b9cc7L119 --- https://github.com/truenas/scale-build/pull/139#discussion_r617370994 )

scale_build/image/iso.py
sonicaj: Keeping the same behavior as original implementation, checking in shell we have
```
Setting up grub2-common (2.04-16) ...
Setting up grub-pc-bin (2.04-16) ...
Setting up shim-signed:amd64 (1.33+15+1533136590.3beb971-7) ...
Setting up grub-efi-amd64 (2.04-16) ...

Creating config file /etc/default/grub with new version
/usr/sbin/grub-probe: error: cannot find a device for / (is /dev mounted?).
dpkg: error processing package grub-efi-amd64 (--configure):
 installed grub-efi-amd64 package post-installation script subprocess returned error exit status 1
dpkg: dependency problems prevent configuration of grub-efi:
 grub-efi depends on grub-efi-amd64 (= 2.04-16); however:
  Package grub-efi-amd64 is not configured yet.

dpkg: error processing package grub-efi (--configure):
 dependency problems - leaving unconfigured
Processing triggers for libc-bin (2.31-9) ...
Processing triggers for shim-signed:amd64 (1.33+15+1533136590.3beb971-7) ...
Errors were encountered while processing:
 grub-efi-amd64
 grub-efi
E: Sub-process /usr/bin/dpkg returned an error code (1)
root@scale:/# echo $?
100
root@scale:/#
```
So it was always failing but apparently our use case was still being fulfilled.

scale_build/package.py
themylogin: Why should caller know about this context manager and not `setup` function itself?

scale_build/utils/paths.py
```diff
@@ -0,0 +1,27 @@
...
+CONF_GRUB = os.path.join(BUILDER_DIR, 'scripts/grub.cfg')
+DPKG_OVERLAY = os.path.join(TMP_DIR, 'dpkg-overlay')
+GIT_MANIFEST_PATH = os.path.join(LOG_DIR, 'GITMANIFEST')
+GIT_LOG_PATH = os.path.join(LOG_DIR, 'git-checkout.log')
+HASH_DIR = os.path.join(TMP_DIR, 'pkghashes')
+MANIFEST = os.path.join(BUILDER_DIR, 'conf/build.manifest')
+PKG_DIR = os.path.join(TMP_DIR, 'pkgdir')
+PKG_LOG_DIR = os.path.join(LOG_DIR, 'packages')
+RELEASE_DIR = os.path.join(TMP_DIR, 'release')
+REQUIRED_RAM = 16  # GB
```
themylogin, line 24: Is this a path?

scale_build/utils/run.py
```diff
@@ -0,0 +1,44 @@
...
+    check = kwargs.pop('check', True)
+    shell = kwargs.pop('shell', False)
+    logger = kwargs.pop('logger', None)
+    env = kwargs.pop('env', None) or os.environ
+    if logger:
+        kwargs['stderr'] = subprocess.STDOUT
+
+    proc = subprocess.Popen(args, stdout=kwargs['stdout'], stderr=kwargs['stderr'], shell=shell, env=env)
+    if logger:
+        for line in map(lambda l: l.rstrip().decode(errors='ignore'), iter(proc.stdout.readline, b'')):
```
themylogin, line 23: Is `stderr` ignored?

themylogin: And can't you just open `stdout` and `stderr` with `encoding` and `errors` by default?

scale_build/utils/logger.py
```diff
@@ -0,0 +1,13 @@
+import logging
+import os
+
+from .paths import LOG_DIR
+
+
+def get_logger(logger_name, logger_path=None, mode='a+'):
```
themylogin, line 7: Where is `logger_path=None` used? What will happen in `os.path.join(LOG_DIR, logger_path)` if `logger_path` is `None`?

scale_build/packages/utils.py
```diff
@@ -0,0 +1,35 @@
...
+        deps.update(
+            get_install_deps(packages, deps, packages[dep].install_dependencies | packages[dep].build_dependencies)
+        )
+    return deps
+
+
+def chroot_run(command, handle=None, **kwargs):
+    if handle:
+        kwargs['stdout'] = kwargs['stderr'] = handle
+    return run(f'chroot {DPKG_OVERLAY} /bin/bash -c "{command}"', shell=True, **kwargs)
```
themylogin, line 35: What will happen if `command` contains `"`?

scale_build/packages/utils.py
```diff
@@ -0,0 +1,35 @@
...
+    for dep in filter(bool, map(str.strip, build_depends_str.split(','))):
+        for subdep in filter(bool, map(str.strip, dep.split('|'))):
+            index = subdep.find('(')
+            if index != -1:
+                subdep = subdep[:index].strip()
+            deps.append(subdep)
+    return deps
+
+
+def get_install_deps(packages, deps, deps_list):
```
themylogin, line 23: Are they `dependencies`, `depends` or `deps`?

scale_build/packages/utils.py
```diff
@@ -0,0 +1,35 @@
...
+    for dep in filter(lambda p: p in packages, deps_list):
+        deps.add(packages[dep].source_name)
+        deps.update(
+            get_install_deps(packages, deps, packages[dep].install_dependencies | packages[dep].build_dependencies)
```
themylogin, line 27: Why are `build_dependencies` included too?

scale_build/packages/package.py
```diff
@@ -0,0 +1,189 @@
...
+    def source_path(self):
+        return os.path.join(SOURCES_DIR, self.name)
+
+    @property
+    def binary_packages(self):
+        if self._binary_packages:
+            return self._binary_packages
+
+        if self.name == 'kernel' or (self.predepscmd and not self.deps_path):
+            # We cannot determine dependency of this package because it does not probably have a control file
```
themylogin, line 86: What do you mean by grub package not having a control file?

themylogin (changes requested):

scale_build/packages/utils.py
sonicaj: If we want to build package `X` and it has package `C` as a build time dependency and package `C` has package `B` as build time dependency and package `A` as install time dependency, we need `A/B/C` in order to build `X`. Why we need `C`'s build time dependency is that because it's not already built to be available in our mirror and we do need to build it.

scale_build/utils/run.py
sonicaj: It's not ignored
```
    if logger:
        kwargs['stderr'] = subprocess.STDOUT
```

sonicaj: Although i think i should separate logger implementation from the default `run` function. Thank you for pointing it out.

scale_build/packages/package.py
themylogin: If we run `predepscmd` (which I suppose are meant to be run before retrieving dependencies), it will appear

scale_build/packages/utils.py
themylogin: Then I suppose `get_install_deps` is a misleading name?

scale_build/utils/run.py
themylogin: You are not consuming it here, it could lead to a deadlock.

themylogin: Sorry, totally missed that part.

Popening with `encoding` and `errors` by default is still a suggestion.

scale_build/package.py
sonicaj: `setup` function cleans up relevant mount points if it succeeds but it does not if it fails. So using it as a context manager ensures that they would be cleaned.

setup.cfg
```diff
@@ -0,0 +1,4 @@
+[flake8]
+max-line-length=120
+per-file-ignores =
+    ./truenas_install/__main__.py:E501
```
themylogin, line 4: Can you just fix the file please?

scale_build/update_image.py
```diff
@@ -0,0 +1,45 @@
...
+    os.makedirs(CHROOT_BASEDIR)
+    logger.debug('Bootstrapping TrueNAS rootfs [UPDATE]')
+
+    package_bootstrap_obj = PackageBootstrapDirectory(update_image_logger)
+    package_bootstrap_obj.setup()
+
+    logger.debug('Installing TrueNAS rootfs package [UPDATE]')
+    setup_chroot_basedir(package_bootstrap_obj, package_bootstrap_obj.logger)
+    install_rootfs_packages(update_image_logger)
+    umount_chroot_basedir()
```
themylogin, line 39: Will we reach `umount_chroot_basedir` if `install_rootfs_packages` raises an exception?

scale_build/update_image.py
```diff
@@ -0,0 +1,45 @@
...
+
+    logger.debug('Building TrueNAS rootfs image [UPDATE]')
+    build_rootfs_image(update_image_logger)
+    umount_tmpfs_and_clean_chroot_dir()
```
themylogin, line 43: Will we reach `umount_tmpfs_and_clean_chroot_dir` if some code above raises an exception?

scale_build/preflight.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def is_root():
+    return os.geteuid() == 0
+
+
+def retrieve_missing_packages():
+    missing = {pkg for pkg in WANTED_PACKAGES if not shutil.which(pkg)}
+    if not os.path.exists('/lib/grub/x86_64-efi') and not os.path.exists('/usr/lib/grub/x86_64-efi'):
+        missing.add('grub-efi-amd64-bin')
```
themylogin, line 29: Do we need grub on build host?

themylogin (changes requested):
Thank you! Very little to fix is left.

truenas_install/__main__.py
```diff
@@ -132,15 +132,17 @@ def install_grub_freebsd(input, manifest, pool_name, dataset_name, disks):
...
+                                f"""$menuentry_id_option 'gnulinux-simple-{fs_uuid}' {{
                 load_video
                 insmod gzio
                 if [ x$grub_platform = xxen ]; then insmod xzio; insmod lzopio; fi
                 insmod part_gpt
                 insmod zfs
                 search --no-floppy --fs-uuid --set=root {fs_uuid}
                 echo	'Loading Linux {manifest['kernel_version']} ...'
-                linux	/ROOT/{manifest['version']}@/boot/vmlinuz-{manifest['kernel_version']} root=ZFS={dataset_name} ro {cmdline} console=tty1 zfs_force=yes
+                linux	/ROOT/{manifest['version']}@/boot/vmlinuz-{manifest['kernel_version']}"""
```
themylogin, line 144: Missing ending whitespace here

scale_build/iso.py
```diff
@@ -0,0 +1,51 @@
...
+
+    logger.debug('Bootstrapping CD chroot [ISO]')
+    cdrom_bootstrap_obj = CdromBootstrapDirectory(iso_logger)
+    cdrom_bootstrap_obj.setup()
+
+    setup_chroot_basedir(cdrom_bootstrap_obj, cdrom_bootstrap_obj.logger)
+
+    logger.debug('Installing packages [ISO]')
+    install_iso_packages(iso_logger)
+    umount_chroot_basedir()
```
themylogin, line 45: I see same cleanup issues here

themylogin (changes requested):

themylogin (approved):
Great, thank you!

PR #142, opened 2021-04-29
NAS-110424 / 21.06 / Validate build manifest

sonicaj:


requirements.txt
```diff
@@ -10,3 +10,4 @@ PyYAML==5.4.1
 requests==2.25.1
 toposort==1.6
 urllib3==1.26.4
+jsonschema==3.2.0
```
themylogin, line 13: This file is generated by `pip freeze` should be in the alphabetical order to avoid meaningless diffs

scripts/validate_manifest.py
```diff
@@ -0,0 +1,159 @@
...
+        with open(manifest_path, 'r') as f:
+            manifest = yaml.safe_load(f.read())
+    except FileNotFoundError:
+        error_str = f'{manifest_path!r} does not exist'
+    except yaml.YAMLError:
+        error_str = f'Unable to read {manifest_path!r} contents. Can you please confirm format is valid ?'
+
+    if error_str:
+        print(f'[\033[91mFAILED\x1B[0m]\t{error_str}')
+        exit(1)
```
themylogin, line 20: Can we also validate manifest when we load it please?

themylogin (changes requested):

Makefile
```diff
@@ -16,13 +16,17 @@ endif
 
 all: checkout packages update iso
 
-clean: check
+clean: validate
```
themylogin, line 19: Is `Makefile` the only allowed way to launch `scale_build`?

themylogin (approved):

PR #154, opened 2021-05-19
NAS-110494 / 21.06 / Make sure machine-id is unique for different scale systems

sonicaj:
This commit fixes 2 issues, it makes sure that systems running same scale version have different machine-ids as that's what is being used by linux to determine unique mac addresses for virtual interfaces. Secondly it makes sure that on upgrades machine-id persists so that the kernel considers it to be the same machine after an upgrade.

### DIFF
--- a/truenas_install/__main__.py
+++ b/truenas_install/__main__.py
@@ -253,12 +253,22 @@ def main():
 
                 write_progress(0.9, "Performing post-install tasks")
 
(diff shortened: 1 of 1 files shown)


yocalebo (approved):
This fixes an old bug I ran into when installing scale on HA systems. The 2 controllers had the same /etc/machine-id so I ended up using /etc/hostid but good to see proper solution.

PR #161, opened 2021-06-06
NAS-110934 / 21.08 / Allow specifying environment variables independently

sonicaj:
This commit adds changes to allow specifying enviornment variables independently for each package. This is helpful to avoid influence of one env variable meant for a specific package on another which might result in changed behavior when it builds.

### DIFF
--- a/scale_build/packages/build.py
+++ b/scale_build/packages/build.py
@@ -19,8 +19,7 @@ class BuildPackageMixin:
             env={
                 **os.environ,
                 **APT_ENV,
-                'CONFIG_DEBUG_INFO': 'Y',  # Build kernel with debug symbols
-                'CONFIG_LOCALVERSION': '+truenas',
+                **self.env,
             }
         )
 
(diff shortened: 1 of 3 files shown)


sonicaj:
ok to test

scale_build/packages/build.py
```diff
@@ -71,7 +70,7 @@ def _build_impl(self):
                 predep_cmd = predep_entry['command']
                 skip_cmd = False
                 for env_var in predep_entry['env_checks']:
-                    if os.environ.get(env_var['key']) != env_var['value']:
+                    if all(f(env_var['key']) != env_var['value'] for f in (os.environ.get, self.env.get)):
```
themylogin, line 73: You forgot `APT_ENV` here.
Can we make a single method that will construct resulting `env` for command and check against it instead?

themylogin (changes requested):

themylogin (approved):

PR #162, opened 2021-06-07
NAS-110826 / 21.08 / Build inadyn from source

sonicaj:
This PR adds changes to build inadyn from source as the package has not been maintained since mid 2010s in debian repository and we are outdated.

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -181,7 +181,7 @@ sources:
   repo: https://github.com/truenas/openseachest
 - name: inadyn
   branch: master
-  repo: https://github.com/troglobit/inadyn.git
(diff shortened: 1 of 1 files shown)


conf/build.manifest
```diff
@@ -177,6 +179,9 @@ sources:
 - name: openseachest
   branch: truenas/master
   repo: https://github.com/truenas/openseachest
+- name: inadyn
+  branch: master
+  repo: https://github.com/troglobit/inadyn.git
```
william-gr, line 184: Can we fork it? We need it for tagging as part of the release engineering process

sonicaj: Definitely, doing it asap

william-gr (approved):

PR #165, opened 2021-07-01
NAS-111272 / 21.08 / Re-install scale-build if there are manual edits

sonicaj:
This commit adds changes to re-install scale-build in virtualenv if there are manual edits, this is very helpful for devs during debugging where they don't have to manually remove virtual env.


rick-mesta (approved):
suh-weet 💓

PR #167, opened 2021-07-13
NAS-111358 / 21.08 / remove python-is-python2 package

yocalebo:
`python2` is being installed on SCALE by upstream Debian. See https://packages.debian.org/sid/main/python-is-python2 for details. We already ensure all the dependent ports (and the ones we write) use python3 so no need to keep this around.

However, I'm explicitly installing the `python-is-python3` package since it, literally, only creates a symlink to `/bin/python`. It would be incredibly time consuming to track every single 3rd party package/script  that references the `python` binary file so this should keep the unforeseen fall-out minimal.

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -98,6 +98,7 @@ base-prune:
 - gstreamer1.0-libav
 - libgdk-pixbuf2.0-0
 - x11-common
+- python-is-python2
 
 #
 # Update build-epoch when you want to force the next build to be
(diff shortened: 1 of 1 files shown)


sonicaj (approved):

william-gr:
Is tere a python-is-python3? ;)

yocalebo:
> Is there a python-is-python3? ;)

Actually there is, it literally just creates a symlink from python3 binary to /usr/bin/python 😄 I think I'm going to install it actually since it's impossible to track down every potential 3rd party program/script blah blah that could be using that binary (`python`). I've updated the comment to this PR and I'll update the PR.

william-gr:
Yup, good call, thats what I was thinking

PR #169, opened 2021-07-19
NAS-110816 / 21.08 / Set bootloader property so zectl can update grub

sonicaj:
This commit adds changes to ensure bootloader property is set for BE's so that zectl is able to consume our grub plugin to update grub after making changes to boot environments.


themylogin (changes requested):
Shouldn't we do this later after everything is installed?

themylogin (approved):

PR #170, opened 2021-07-19
Allow upgrading FreeBSD installed on bios-boot partition

themylogin:


truenas_install/__main__.py
```diff
@@ -151,8 +151,9 @@ def install_grub_freebsd(input, manifest, pool_name, dataset_name, disks):
     run_command(["grub-mkconfig", "-o", "/boot/grub/grub.cfg"])
 
     for disk in disks:
-        if boot_partition_type == "freebsd-boot":
-            run_command(["gpart", "modify", "-i", "1", "-t", "bios-boot", f"/dev/{disk}"])
+        if boot_partition_type in ["bios-boot", "freebsd-boot"]:
+            if boot_partition_type != "bios-boot":
```
sonicaj, line 155: Maybe we should do `boot_partition_type == "freebsd-boot"` ?

themylogin: It should be `!= "bios-boot"` as `bios-boot` is the only type of partition for which the following command is not necessary (and will raise an error)

sonicaj: Yes but there are only 2 partition types which will ever come to this conditional because of L154 ?

themylogin: @sonicaj it will when we add 3rd partition type

sonicaj (approved):

PR #175, opened 2021-08-19
NAS-111886 / 21.10 / Do not automatically remove tmp/sources directory with epoch changes

sonicaj:
This commit adds changes to not remove tmp/sources directory automatically when epoch changes but instead raise an exception. In case an env variable is set, system will then force the removal of these directories otherwise they would have to be done manually.


scale_build/config.py
```diff
@@ -9,6 +9,7 @@
 BUILDER_DIR = os.getenv('BUILDER_DIR', './')
 BRANCH_OUT_NAME = os.getenv('NEW_BRANCH_NAME')
 BRANCH_OVERRIDES = {k[:-(len('_OVERRIDE'))]: v for k, v in os.environ.items() if k.endswith('_OVERRIDE')}
+FORCE_CLEANUP_WITH_EPOCH_CHANGE = os.getenv('FORCE_CLEANUP_WITH_EPOCH_CHANGE', False)
```
themylogin, line 12: Environment variables are strings. Here it will be either string or bool. And `FORCE_CLEANUP_WITH_EPOCH_CHANGE=0` environment variable will be treated as `True` because `bool('0')` is `True`.

One of the correct ways to read bool environment variable is `bool(int(os.getenv('FORCE_CLEANUP_WITH_EPOCH_CHANGE') or '0'))`

Same applies for `PKG_DEBUG` below.

themylogin (changes requested):

scale_build/config.py
sonicaj: Adjusting changes to `PKG_DEBUG` in the related PR :)

themylogin (approved):

PR #177, opened 2021-08-19
DOCS-2458 / Added code to artifacts debug packages in all build piplines

ericbsd:
I also took time to fix spaces in all jenkins piplines

### DIFF
--- a/jenkins/FullBuild
+++ b/jenkins/FullBuild
@@ -46,9 +46,9 @@ pipeline {
     stage('Artifact') {
       steps {
         sh 'rm -rf artifacts || true'
-        sh 'mkdir -p artifacts/pkg_dbg'
+        sh 'mkdir -p artifacts/packages'
         sh 'mv src/tmp/release/* artifacts/'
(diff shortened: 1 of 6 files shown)


jenkins/FullBuild
```diff
@@ -46,9 +46,10 @@ pipeline {
     stage('Artifact') {
       steps {
         sh 'rm -rf artifacts || true'
-        sh 'mkdir -p artifacts'
+        sh 'mkdir -p artifacts/pkg_dbg'
         sh 'mv src/tmp/release/* artifacts/'
-        }
+        sh 'cp src/tmp/pkgdir/*dbg* artifacts/pkg_dbg/'
```
sonicaj, line 51: I think it would be useful to download the complete pkg directory as well in case someone wants to just test the new package instead of trying out the complete build ?

anodos325: Oh I like that idea.

ericbsd: OK, let's go for it.

ericbsd: done

anodos325 (approved):

sonicaj (approved):

PR #181, opened 2021-09-03
NAS-112161 / 21.10 / Add python3-pip to scale build

sonicaj:


pkgdemon (approved):

themylogin:
@pkgdemon were there any issues installing PIP from the Jenkins? It would be very convenient to have it in our image for me too, but SCALE is already too big, we should be decreasing its size instead of increasing furthermore.

PR #186, opened 2021-09-26
CodeQL

dustinwloring1988:
CodeQL, Prettier, grettings


william-gr:
What is the intent of this?

freqlabs:
It looks like it automatically commits code cleanup based on some linter? Doesn't seem like a great idea personally. I like clean code, but giving commit access to some random Docker container not so much.

william-gr:
Oh if thats the case then its a big no no. I would be fine if it would just flag anything and not actually try to commit anything.

PR #205, opened 2021-12-16
NAS-113855 / 22.02 / add zstd to SCALE

yocalebo:
`zstd` brings in `unzstd` which allows uncompressing core dump files on SCALE (they're compressed using zstd algorithm by default).

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -167,6 +167,8 @@ additional-packages:
   comment: NAS-111358 (symlinks /usr/bin/python to python3)
 - package: python3-pip
   comment: NAS-112161 ( requested by QE team )
+- package: zstd
+  comment: NAS-113855 ( requested by engineering team )
 
 #
 # List of additional packages installed into TrueNAS SCALE ISO file


themylogin (commented):
Do we really need all this stuff in the ISO?

Can we make a metapackage like `truenas-devel` or `truenas-qe` that will install necessary dependencies?

yocalebo:
I don't know how to do that 😄

themylogin:
I can do it

@rick-mesta what do you think about shipping ISO and update images that contain tools only needed for our internal use? Why do other departments require us to ship them, why can't they just use apt to install them?

sonicaj:
I do think that we should probably investigate and see if we can reduce the size of SCALE iso and for tools which can be easily downloaded with apt, we should probably not ship them by default :)

yocalebo:
Like the nvidia packages!

sonicaj:
They are used to automatically setup/configure GPU - so are not really only needed for our internal use :)

yocalebo:
detecting the GPU and then downloading the necessary packages would be my vote. (It's also what every other linux distrubution does as well)

yocalebo:
Anyways @themylogin , I'll close this PR and then pass the ticket to you if that's okay?

themylogin:
@yocalebo let's hear our leader's @rick-mesta opinion first. There might be reasons to leave this as-is.

rick-mesta:
sorry gents... been busy w/other goodies: I can see a case for both, actually. I'm in favor of reducing the iso image to be sure, however, we have to temper that with how much time will be spent on tracking down the right tools every time we debug something. @kmoore134 or @william-gr may have more insight as to the actual number of steps in order to set up the right amount of tools vs. iso image size

william-gr (approved):
Reducing the image size is not a goal for Angelfish this late in the game.
We can revisit this for Bluefin ;)

cTurtle98:
I am on TrueNAS-SCALE-22.02.3 and when I try to run unzstd i get command not found.

themylogin:
@cTurtle98 it's not in the image anymore, you have to install it manually

PR #207, opened 2021-12-20
NAS-113907 / 22.02 / Build with Ganesha v4.0

anodos325:


conf/build.manifest
```diff
@@ -260,6 +263,12 @@ sources:
   batch_priority: 0
   explicit_deps:
     - openzfs
+- name: nfs-ganesha
+  repo: https://github.com/truenas/nfs-ganesha
+  branch: SCALE-V4.0
+  predepscmd:
+    - "git submodule update --init --recursive"
```
sonicaj, line 270: This should be happening when we run the checkout target, can you please confirm if it's not happening ?

anodos325: Oh interesting. I wasn't aware we are doing that. Do we detect when there's an update in a submodule that isn't reflected in the version cached in builder?

sonicaj: When we run `checkout` target, it should automatically update the repository if it's already checked out or otherwise it should clone it.

sonicaj (approved):

PR #209, opened 2022-01-02
NAS-113649 / 22.02 / Enable walinuxagent as required on upgrades

sonicaj:
This commit adds changes to enable walinuxagent in the new BE if system in question currently is azure.


yocalebo (commented):
Is the wrapper method really necessary?

sonicaj:
@yocalebo seemed neater to me as the existing method has a specific use case. If you have concerns, please let me know and i can remove/adjust it as required ?

themylogin (approved):

PR #213, opened 2022-01-12
Have waagent install as a dep for truenas

sonicaj:
We want to install waagent before truenas as waagent is not disabled otherwise because it's not installed at the time. So we will be adding waagent as a dependency for truenas.


themylogin:
Why not simply add it to `truenas` package dependencies?

themylogin (approved):

PR #215, opened 2022-01-20
NAS-114149 / 22.02 / Proper way to mount boot media

themylogin:


themylogin:
time 8:00

yocalebo:
This broke mounting the ISO file on install.

PR #221, opened 2022-01-30
NAS-114205 / 22.02 / Have empty `/boot/grub` directory in rootfs image

themylogin:


scale_build/image/update.py
```diff
@@ -104,6 +104,8 @@ def post_rootfs_setup():
 def custom_rootfs_setup():
     # Any kind of custom mangling of the built rootfs image can exist here
 
+    os.makedirs(os.path.join(CHROOT_BASEDIR, '/boot/grub'), exist_ok=True)
```
sonicaj, line 107: We perhaps want to remove leading `/` ?
```
os.path.join('/mna/tabc', '/qe/se')
'/qe/se'
```

william-gr: Seems like we are not testing the PRs :)

themylogin: @william-gr I don't have scale builder locally and push PRs to use Jenkins one. I forgot to mark this as WIP, sorry.

sonicaj (approved):

PR #223, opened 2022-02-02
Failing to enable services from our JSON file should not be fatal

anodos325:
This file may have outdated info in it. Worst case scenario, user ends up with some unexpectedly enabled services.

### DIFF
--- a/truenas_install/__main__.py
+++ b/truenas_install/__main__.py
@@ -84,7 +84,7 @@ def enable_user_services(root, old_root):
 
     if systemd_units:
         try:
-            run_command(["chroot", root, "systemctl", "enable"] + systemd_units)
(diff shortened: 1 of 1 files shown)


kmoore134 (approved):

themylogin (changes requested):
Cool, but can we please use `check=False`?

truenas_install/__main__.py
```diff
@@ -83,7 +83,10 @@ def enable_user_services(root, old_root):
         ]
 
     if systemd_units:
-        run_command(["chroot", root, "systemctl", "enable"] + systemd_units)
+        try:
+            run_command(["chroot", root, "systemctl", "enable"] + systemd_units, check=False)
```
william-gr, line 87: I think he meant that with check=False we dont need try/except?

PR #238, opened 2022-02-19
NAS-114929 / 22.02.1 / add py-nvme to build

yocalebo:
This module adds support persistent reservations for nvme devices.

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -388,6 +388,9 @@ sources:
 - name: py_fenced
   repo: https://github.com/truenas/py-fenced
   branch: master
+- name: py_nvme
+  repo: https://github.com/truenas/py-nvme
+  branch: master
 - name: swagger
   repo: https://github.com/truenas/swagger
   branch: main


william-gr (commented):
Building failing, does it need to be added as a dependency of middlewared?

sonicaj:
It should probably be a dependency of `py-fenced`

yocalebo:
retest this please

yocalebo:
Yeah, not sure. python3-nvme is a predepend for python3-fenced and those are building without issues. Not sure why middlewared is failing.

sonicaj:
It should be added as a `Depends` for `python3-fenced` as well

yocalebo:
> It should be added as a `Depends` for `python3-fenced` as well

I've only ever listed `python3-sgio` as a pre-depend for `python3-fenced` so why would this be different?

sonicaj:
@yocalebo because it's already added as `Depends` in middlewared control file. Also idk why you are using `Pre-Depends` :) ( https://www.debian.org/doc/debian-policy/ch-relationships.html )
```
Pre-Depends should be used sparingly, preferably only by packages whose premature upgrade or installation would hamper the ability of the system to continue with any upgrade that might be in progress.
```

yocalebo:
> @yocalebo because it's already added as `Depends` in middlewared control file. Also idk why you are using `Pre-Depends` :) ( https://www.debian.org/doc/debian-policy/ch-relationships.html )
> 
> ```
> Pre-Depends should be used sparingly, preferably only by packages whose premature upgrade or installation would hamper the ability of the system to continue with any upgrade that might be in progress.
> ```

Because that's how it has always been 😄 Why change something when it has been working? I added `python3-nvme` as a build depend here: https://github.com/truenas/middleware/pull/8319

sonicaj:
We should correctly specify the dependency right now then ;)

yocalebo:
retest this please

yocalebo:
@sonicaj ping since the surrounding issues have been fixed.

sonicaj (approved):

PR #239, opened 2022-02-23
NAS-114949 / 22.02.1 / Update scale builder

sonicaj:
This PR adds the following changes:

1. Update builder to point to new apt mirrors
2. Make sure we update version appropriately
3. Make sure nightly train is pointing to correct release


scale_build/config.py
```diff
@@ -22,4 +22,4 @@
 if os.getenv('TRUENAS_VERSION'):
     VERSION = os.getenv('TRUENAS_VERSION')
 else:
-    VERSION = f'22.02.1-MASTER-{BUILD_TIME_OBJ.strftime("%Y%m%d-%H%M%S")}'
+    VERSION = f'22.12-MASTER-{BUILD_TIME_OBJ.strftime("%Y%m%d-%H%M%S")}'
```
william-gr, line 25: @themylogin Will we be able to upgrade to 22.12.0 when it releases if you are running 22.12-MASTER-20220720?

themylogin: @william-gr 
```
root@truenas[~]# midclt call system.version
TrueNAS-SCALE-22.02.0
root@truenas[~]# python3
Python 3.9.2 (default, Feb 28 2021, 17:03:44) 
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from middlewared.plugins.update_.utils import can_update
>>> can_update("22.02.1-MASTER-2022", "22.12-MASTER-2022")
True
```

themylogin (commented):
Can we just substitute `{codename}` in repo URLs?

sonicaj:
We should definitely do that and also i think we should probably optimize the versioning bit - it should be in manifest until an env variable overrides it. Is it okay if we do that in a separate PR ?

PR #243, opened 2022-03-02
NAS-114800 / 22.12 / Add truenas docker to build manifest

Qubad786:


conf/build.manifest
```diff
@@ -216,6 +219,10 @@ iso-packages:
 # to be built
 ############################################################################
 sources:
+- name: docker
+  repo: https://github.com/truenas/moby
+  branch: truenas/20.10
```
william-gr, line 224: Why 20.10?

Qubad786: @william-gr  [truenas/20.10](https://github.com/truenas/moby/tree/truenas/20.10) is in-sync with upstream 20.10 (+ having our additional updates). That's the version, we're using right now.

PR #245, opened 2022-03-04
NAS-115069 / 22.12 / Destroy old `boot-pool/grub` dataset when upgrading FreeBSD installat…

themylogin:
…ions recursively because it can contain snapshots


themylogin:
time 0:20

yocalebo (approved):

PR #251, opened 2022-03-15
NAS-115287 / 22.12 / Provide kernel version as an environment variable to package builds

sonicaj:


william-gr:
Only one question, what is the intent of this? Why would you need an env var for the kernel version.

sonicaj:
Motivation behind it is to handle packages like `scst` which install linux headers which require the kernel version in the name iirc and this means each time we update kernel version, `scst` needs to be updated manually. With this available in the env, a predepscmd can be specified which would help us avoid commits to `scst` each time kernel is updated.

themylogin (changes requested):
Can/should we do one of these things?
a) Retrieve `kernel_version` automatically during the build
b) Assert that `kernel_version` matches the kernel version we are using and fail the build if they don't match
Otherwise, someone might forgot to update the build manifest with corresponding consequences.

sonicaj:
After talking with @yocalebo, let's fix the usage in scst as required.

PR #253, opened 2022-03-15
NAS-115260 / 22.12 / Add Infiniband/Ethernet connetcx adapters

sonicaj:


conf/build.manifest
```diff
@@ -189,6 +189,12 @@ additional-packages:
   comment: NAS-114723
 - package: powertop
   comment: requested by community (NAS-113898)
+- package: rdma-core
+  comment: requested by community (NAS-115260)
+- package: ibutils
+  comment: requested by community (NAS-115260)
+- package: infiniband-diags
+  comment: requested by community (NAS-115260)
```
william-gr, line 197: Just to make sure, have we verified we are free to distribute those? e.g. not non-free

yocalebo:
We should probably consider our enterprise product line and whether or not adding these packages will effect us in the long term.

conf/build.manifest
sonicaj: Yes, they are non-free. I verified that via debian tracker

william-gr:
@yocalebo can you elaborate? i admittedly dont know enough about infiniband

sonicaj:
@yocalebo can you have a peek at @william-gr's comment please ;)

sonicaj:
ping @yocalebo

yocalebo:
The problem that I have with this is that we do not support infiniband/rdma on SCALE and we're just adding this into the product so a user has the convenience of not having to run 1 more CLI argument. (`apt update; apt install <insert-pkgs-here>`)

Furthermore, we only build the `IPoIB` driver on CORE which doesn't even really make sense by itself.

sonicaj:
@william-gr should we still have these based on @yocalebo suggestion ? We do have apt disabled though for context

william-gr:
I agree we don't officially support those, but I also don't see why not just add the packages and let the users deal with it themselves.

Unless these packages are big ,pose any risk or infringe license redistribution somehow?

yocalebo:
> I agree we don't officially support those, but I also don't see why not just add the packages and let the users deal with it themselves.
> 
> Unless these packages are big ,pose any risk or infringe license redistribution somehow?

Installed size for all of these are ~3ish MB so not too big. Since the user has to go to the CLI to configure anything related to rdma/infiniband anyways, I don't see why it's that much of a problem for the user to type in a few more CLI commands to get them installed?

william-gr:
It doesn't survive reboots and as Waqar said apt is readonly. I also have plans we should make root filesystem readonly.

yocalebo:
Well, the 3 immediate packages installed are around 3MB but I have no idea how much more space is needed for the dependencies 😄

yocalebo:
> It doesn't survive reboots and as Waqar said apt is readonly. I also have plans we should make root filesystem readonly.

Okay, but then if we move to readonly and lock down apt and actively tell users that we don't support configuration via the CLI, then why would we add these utilities for a user to do exactly what we don't want them to do?

william-gr:
@yocalebo @sonicaj I talked to JT about it as well. In the end I agree with Caleb, there is probably little sense in adding it at this time.

sonicaj:
Thanks for confirming @william-gr @yocalebo

PR #255, opened 2022-03-17
Better log any exceptions when building packages

sonicaj:


yocalebo (approved):

themylogin (approved):

themylogin (commented):
Why not `exc_info=True` with full traceback?

PR #261, opened 2022-03-24
Switch angelfish back to building from SCALE-v4-15-stable

anodos325:
I will soon have SCALE-v4-16-stable ready for BlueFin. We don't
need extra branches being maintained for angelfish.

TrueNAS 13 / SCALE Angelfish are based on SCALE-v4-15-stable.

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -272,7 +272,7 @@ sources:
   generate_version: false
 - name: truenas_samba
   repo: https://github.com/truenas/samba
-  branch: stable/angelfish
+  branch: SCALE-v4-15-stable
   generate_version: false
   batch_priority: 0
   explicit_deps:


anodos325:
I didn't realize that someone had added a new branch to Samba repo and made it default for angelfish builds. I don't see why we need a separate branch for this.

yocalebo (approved):

rick-mesta (approved):

william-gr:
Must be something release engineering did. Might be worth asking them if its necessary for anything in their system.

william-gr:
Actually, Waqar reminded me we did this when branching out for Angelfish, so all the repos have the same branch for consistency sake.
IMHO its simpler and safer to do that, so we dont get confused on what goes to Angelfish, which is supposed to be stable now, without having to map in the build repo.

PR #262, opened 2022-03-24
NAS-115468 / 22.12 / parallelize the git checkout of repo sources

yocalebo:
Parallelize the git checkout of repos. This greatly reduces the time needed to checkout the repos since it's currently being done synchronously. The major outlier is the `github.com/truenas/linux` repo for obvious reasons, but now all the other repos will be checked out without having to wait for that one to finish.

### DIFF
--- a/scale_build/checkout.py
+++ b/scale_build/checkout.py
@@ -1,16 +1,37 @@
 import logging
+from concurrent.futures import ThreadPoolExecutor, as_completed
 
 from .utils.git_utils import retrieve_git_remote_and_sha, update_git_manifest
 from .utils.package import get_packages
 
 
 logger = logging.getLogger(__name__)
+MAX_THREADS = 10  # number of threads to spin up to checkout/pull github sources
 
 
 def checkout_sources():
     info = retrieve_git_remote_and_sha('.')
     update_git_manifest(info['url'], info['sha'], 'w')
-    logger.info('Starting checkout of source')
 
-    for package in get_packages():
-        package.checkout(package.get_branch_override())
+    pkgs = {
+        pkg.pkg_name: {
+            'checkout_method': pkg.checkout,
+            'get_branch_override_method': pkg.get_branch_override,
+            'branch_override': None,
+        } for pkg in get_packages()
+    }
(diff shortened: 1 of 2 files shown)


scale_build/checkout.py
```diff
@@ -1,16 +1,37 @@
...
+            try:
+                branch_override = fut.result()
+            except Exception:
+                logger.warning('Failed to generate branch override for %r', pkg_name, exc_info=True)
+                branch_override = None
+
+            pkgs[pkg_name]['branch_override'] = branch_override
+
+        logger.info('Starting checkout of sources')
+        [exc.submit(v['checkout_method'], v['branch_override']) for pkg, v in pkgs.items()]
```
sonicaj, line 37: Can we please perhaps properly wait for the futures to complete rather then relying on the context manager to make sure it exits when the tasks have completed in case we make some more edits down the road and forget that we are relying on the context manager to do the waiting and not being explicit about it ?
```
with ThreadPoolExecutor(max_workers=3) as exc:
        futures = [exc.submit(wait, f'thread_{i}') for i in range(10)]
        wait(futures, return_when=ALL_COMPLETED)
```

scale_build/checkout.py
```diff
@@ -1,16 +1,37 @@
...
+    }
+    with ThreadPoolExecutor(max_workers=MAX_THREADS) as exc:
+        logger.info('Getting override for branches')
+        branchoverrides_to_pkgs = {exc.submit(v['get_branch_override_method']): k for k, v in pkgs.items()}
+        for fut in as_completed(branchoverrides_to_pkgs):
+            pkg_name = branchoverrides_to_pkgs[fut]
+            try:
+                branch_override = fut.result()
+            except Exception:
+                logger.warning('Failed to generate branch override for %r', pkg_name, exc_info=True)
```
sonicaj, line 31: I am not sure if we should just move on if this fails for some reason. I think we should break and raise an appropriate exception that x number of packages were not able to determine branch override as this can result in automated jobs continuing even when we have errors and publishing from wrong branches.

sonicaj (approved):
Thanks!

PR #266, opened 2022-04-01
NAS-115605 / 22.12 / Fix ValueError crashes

yocalebo:
I'm trying to work on some other tickets and I got a message to set `SKIP_SOURCE_REPO_VALIDATION` so I did `export SKIP_SOURCE_REPO_VALIDATION=y`

Tried to do `make packages` and it crashed with `ValueError` because we're trying to cast the result to `int`

With these changes, I've added a `get_env_variable` function that properly sets the environment variables in a programmatic way that should, hopefully, prevent these types of crashes in the future.

### DIFF
--- a/scale_build/config.py
+++ b/scale_build/config.py
@@ -2,6 +2,8 @@ from os import environ, cpu_count
 from time import time
 from datetime import datetime
 
+_VERS = '22.12-MASTER'
+
 
 def get_env_variable(key, default_value, _type):
     value = environ.get(key)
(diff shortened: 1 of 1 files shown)


sonicaj (commented):
LGTM, however can we document this in the readme please as specifying a 'N/n/0' would result in this being set still so we can just be explicit about it in the readme

yocalebo:
@sonicaj ping

sonicaj (approved):

PR #269, opened 2022-04-03
Have bullseye-backport mirror for angelfish

sonicaj:
We would like to use our fork of docker for stable/angelfish but that requires golang 1.17. Also based on some reports we are seeing, it might be better to update k3s as well which would also require the newer golang.


conf/build.manifest
```diff
@@ -156,6 +159,9 @@ apt_preferences:
 - Package: "*polkit*"
   Pin: "release n=bullseye-security"
   Pin-Priority: 1000
+- Package: "*golang*"
```
themylogin, line 162: Can we try to keep this list alphabetical please?

themylogin (approved):

conf/build.manifest
sonicaj: Sure, created a ticket for it.

PR #282, opened 2022-04-07
NAS-115619 / 22.12 / Log git checkout of each package separately

Qubad786:
## Description
Moving git logs to separate log files for different packages as packages are being checked out in parallel.

### History
Previously all logs from git were going to same log file in parallel that was making them unreadable.


sonicaj (approved):

Qubad786:
backport

PR #298, opened 2022-05-12
NAS-116170 / 22.12 / retry git commands on failure

yocalebo:
Our SCALE VM builders are failing to checkout git repos. The repo it fails on and when it fails are completely random. Until the root cause can be diagnosed and subsequently fixed, this will retry the update or checkout phase if a failure is encountered.

I've updated the logging verbiage so that we know when a retry is taking place.

### DIFF
--- a/scale_build/packages/git.py
+++ b/scale_build/packages/git.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import shutil
+import contextlib
 
 from scale_build.config import BRANCH_OVERRIDES, TRUENAS_BRANCH_OVERRIDE, TRY_BRANCH_OVERRIDE
 from scale_build.exceptions import CallError
@@ -11,7 +12,6 @@ from scale_build.utils.git_utils import (
 from scale_build.utils.logger import LoggingContext
 from scale_build.utils.paths import GIT_LOG_DIR_NAME, GIT_LOG_DIR
 from scale_build.utils.run import run
-
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


scale_build/packages/git.py
```diff
@@ -45,25 +45,70 @@ def git_log_file(self):
...
+                ['git', '-C', self.source_path, 'checkout', branch],
+            )
+
+        # We're doing retries here because at the time of writing this the iX network
+        # is having issues with an external hop through the routing of the interwebz
+        # getting to github.com. They've found a particular hop is dropping significant
+        # amounts of packets (~75%+). This is happening network wide so we've got the
+        # retries.
+        # NOTE: when the issue is fixed, we could remove this retry logic
+        retries = 3 if retries <= 0 or retries > 10 else retries
```
sonicaj, line 70: Do we need this ? And if we do, do you think it's better to raise an exception that less then 0 or bigger then 10 is not allowed rather then implicitly defaulting to 3 ?

scale_build/packages/git.py
```diff
@@ -45,25 +45,70 @@ def git_log_file(self):
...
+                err = f'Failed cmd {failed[0]!r} with error {failed[1]!r} with returncode {failed[2]!r}.'
+                err += f' Check {failed_log_file!r} for details.'
+                shutil.copyfile(self.git_log_file, failed_log_file)
+                if i == retries:
+                    raise CallError(err)
+                else:
+                    logger.warning(err)
+                    continue
+            else:
+                break
```
sonicaj, line 107: This is not required

scale_build/packages/git.py
```diff
@@ -45,25 +45,70 @@ def git_log_file(self):
...
-                run(['git', 'clone', '--recurse', self.origin, self.source_path])
-                run(['git', '-C', self.source_path, 'checkout', branch])
+                for cmd in cmds:
+                    cp = run(cmd, check=False)
+                    if cp.returncode:
+                        failed = (f'{" ".join(cmd)}', f'{cp.stderr}', f'{cp.returncode}')
+                        break
+
+            if failed:
+                failed_log_file = self.git_log_file + f'.failed.{i}'
```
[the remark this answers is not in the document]
yocalebo: Good idea

sonicaj (approved):
Thanks!

PR #304, opened 2022-06-02
NAS-115611 / 22.12 / NAS-115611: Create EFI boot entries for all boot disks

usaleem-ix:
Currently, grub-install only creates one EFI boot entry for one boot
disk. If that boot disk gets compromised, system cannot boot from any
other disk since the boot entry is not present. Creating boot entries
for all boot disks fixes this issue.

### DIFF
--- a/truenas_install/__main__.py
+++ b/truenas_install/__main__.py
@@ -425,7 +425,7 @@ def main():
                                             os.unlink(os.path.join("/sys/firmware/efi/efivars", item))
 
                             os.makedirs(f"{root}/boot/efi", exist_ok=True)
-                            for disk in disks:
+                            for i, disk in enumerate(disks):
                                 install_grub_i386 = True
                                 efi_partition_number = 2
                                 format_efi_partition = True
@@ -465,14 +465,20 @@ def main():
                                                 "--efi-directory=/boot/efi",
                                                 "--bootloader-id=debian",
                                                 "--recheck",
-                                                "--no-floppy"]
-                                    if not os.path.exists("/sys/firmware/efi"):
-                                        grub_cmd.append("--no-nvram")
+                                                "--no-floppy",
+                                                "--no-nvram"]
                                     run_command(grub_cmd)
 
                                     run_command(["chroot", root, "mkdir", "-p", "/boot/efi/EFI/boot"])
                                     run_command(["chroot", root, "cp", "/boot/efi/EFI/debian/grubx64.efi",
                                                  "/boot/efi/EFI/boot/bootx64.efi"])
+
+                                    if os.path.exists("/sys/firmware/efi"):
+                                        run_command(["chroot", root, "efibootmgr", "-c",
+                                                     "-d", f"/dev/{disk}",
+                                                     "-p", f"{efi_partition_number}",
+                                                     "-L", f"TrueNAS-{i}",
+                                                     "-l", "/EFI/debian/grubx64.efi"])
                                 finally:
                                     run_command(["chroot", root, "umount", "/boot/efi"])
                     finally:


truenas_install/__main__.py
```diff
@@ -465,14 +465,20 @@ def main():
...
 
                                     run_command(["chroot", root, "mkdir", "-p", "/boot/efi/EFI/boot"])
                                     run_command(["chroot", root, "cp", "/boot/efi/EFI/debian/grubx64.efi",
                                                  "/boot/efi/EFI/boot/bootx64.efi"])
+
+                                    if os.path.exists("/sys/firmware/efi"):
+                                        run_command(["chroot", root, "efibootmgr", "-c",
+                                                     "-d", f"/dev/{disk}",
+                                                     "-p", "2",
+                                                     "-L", f"debian-{disk}",
```
freqlabs, line 480: Let's make the label say `TrueNAS` instead of `debian`.

freqlabs (changes requested):

truenas_install/__main__.py
usaleem-ix: Updated.

truenas_install/__main__.py
```diff
@@ -465,14 +465,20 @@ def main():
...
                                     run_command(grub_cmd)
 
                                     run_command(["chroot", root, "mkdir", "-p", "/boot/efi/EFI/boot"])
                                     run_command(["chroot", root, "cp", "/boot/efi/EFI/debian/grubx64.efi",
                                                  "/boot/efi/EFI/boot/bootx64.efi"])
+
+                                    if os.path.exists("/sys/firmware/efi"):
+                                        run_command(["chroot", root, "efibootmgr", "-c",
+                                                     "-d", f"/dev/{disk}",
+                                                     "-p", "2",
```
themylogin, line 479: Is this EFI partition number? Above we have `efi_partition_number` which can also be `1` if we are upgrading a FreeBSD installation.

usaleem-ix: Yes, this is EFI partition number. I will update accordingly

themylogin (approved):
Thank you!

freqlabs (approved):

truenas_install/__main__.py
```diff
@@ -465,14 +465,20 @@ def main():
...
 
                                     run_command(["chroot", root, "mkdir", "-p", "/boot/efi/EFI/boot"])
                                     run_command(["chroot", root, "cp", "/boot/efi/EFI/debian/grubx64.efi",
                                                  "/boot/efi/EFI/boot/bootx64.efi"])
+
+                                    if os.path.exists("/sys/firmware/efi"):
+                                        run_command(["chroot", root, "efibootmgr", "-c",
+                                                     "-d", f"/dev/{disk}",
+                                                     "-p", f"{efi_partition_number}",
+                                                     "-L", f"TrueNAS-{disk}",
```
amotin, line 480: Is it a good idea to put disk name, that may change into the label?  Will it add value or confusion?

usaleem-ix: I added disk name into label to clarify which boot entry belongs to which disk. Having same Label strings is discouraged and `efibootmgr` issues a couple of warnings.

freqlabs: We could to `for i, disk in reversed(enumerate(disks)):` to get a decreasing counter `i` to use in the label, that way the last boot entry we add is 0.

By the way, if middleware has something for adding or replacing boot disks after install we should make sure it does the same thing there as well.

usaleem-ix: I have removed the disk name from label and included a counter. Let me know if we need to go reverse about it.

amotin (approved):

amotin (approved):

freqlabs (approved):

PR #311, opened 2022-06-21
NAS-116685 / 22.12 / Add update file size to manifest

Qubad786:
## Context

It was requested that we add size of the update file to the build manifest so that it can be used interactively by UI to determine if we have enough space to download the update file.


anodos325 (approved):

anodos325:
backport

PR #319, opened 2022-07-25
NAS-116521 / 22.12 / Add wheel group

Qubad786:
## Context

It was requested that we add a `wheel` group to help users migrating from CORE to SCALE making sure their workflow is not affected as in SCALE the group with same uid is known as `root` instead of `wheel`.


yocalebo:
Ping @truenas/middleware

anodos325 (approved):

PR #322, opened 2022-08-05
NAS-114289 / 22.12 / Update serial setup in Installer's grub.cfg

usaleem-ix:
This commit fixes SCALE boot issue in BIOS mode on X series hardware.
In BIOS mode, we were not able to see any output from GRUB on IPMI
since serial setup was not being done properly in SCALE installer's
grub.cfg.

Moreover, currently both boot options provide serial console options
for SCALE, due to which we always see Linux console on serial since
serial settings are saved by installer. User should only select
(115200 baud) boot option if they want serial console, otherwise serial
console should not be provided.

Signed-off-by: Umer Saleem <usaleem@ixsystems.com>


freqlabs:
Why the sleep?

usaleem-ix:
> Why the sleep?

With serial input enabled, sleep helps consume buffered input. `grub.cfg` generated by `grub-install` also includes the sleep for this purpose.

freqlabs (approved):

scripts/grub.cfg
```diff
@@ -16,18 +16,20 @@ if loadfont $prefix/fonts/unicode.pf2 ; then
...
   set lang=en_US
   set gfxpayload=keep
 fi
-terminal_output gfxterm
 
-serial --unit=0 --speed=9600 --word=8 --parity=no --stop=1
+serial --unit=0 --speed=115200 --word=8 --parity=no --stop=1
+terminal_output console serial
+terminal_input console serial
+sleep -i 1
```
william-gr, line 28: What do you think of adding that comment here, the one you replied to Ryan, please?

Sounds useful enough so the next person doesn't get lost :D

usaleem-ix: Sure, I have added the comment.

william-gr: Thank you!

PR #323, opened 2022-08-05
NAS-117519 / 22.12 / Various improvements to builder

sonicaj:
This PR introduces following changes:

1. Conditional builds of sources
2. Allowing ssh based origin of sources
3. Integrating corssl in TrueNAS for FIPS compliance
4. Various minor bug fixes and enhancements in the builder

Motivation behind the above changes is to basically allow building a separate version of TrueNAS SCALE which is FIPS compliant using CorSSL.

For consumers wishing to build FIPS compliant SCALE, they can set the env variable `FIPS=1` when checking out / building packages. Usage would look something like the following:

```
env corssl_OVERRIDE_IDENTITY_FILE_PATH=identities/key FIPS=1 sh -c "make checkout && make packages"
```

`corssl_OVERRIDE_IDENTITY_FILE_PATH` env variable is provided in order for `checkout` to succeed as `corssl` uses SSH based origin and an identity file must be provided either in the manifest or an environment variable so the build system can use that for checking out the repository in question.

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -1,6 +1,7 @@
 ---
 code_name: "Bluefin"
 debian_release: "bullseye"
+identity_file_path_default: "~/.ssh/id_rsa"
 #
 # List of apt repositories that are used and setup inside the build environment
 # for TrueNAS SCALE. These are used to pull additional packages or depend packages
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -248,6 +248,10 @@ iso-packages:
 # to be built
 ############################################################################
 sources:
+- name: corssl
+  repo: git@github.com:iXsystems/corssl.git
+  branch: master
+  generate_version: false
 - name: docker
   repo: https://github.com/truenas/moby
   branch: truenas/20.10
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -258,6 +258,10 @@ sources:
   repo: git@github.com:iXsystems/corssl.git
   branch: master
   generate_version: false
+  build_constraints:
+    - name: FIPS
+      value: true
+      type: boolean
 - name: docker
   repo: https://github.com/truenas/moby
   branch: truenas/20.10
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -154,12 +154,18 @@ apt_preferences:
 - Package: "librrd*"
   Pin: "version 1.99*"
   Pin-Priority: 950
+- Package: "libssl1.1"
+  Pin:  "origin \"\""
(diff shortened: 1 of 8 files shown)


yocalebo:
If we set `FIPS=1`, why are we requiring a path to the key file? The default location is `~/.ssh/id_rsa` and git will use that file by default. It's nice that you added the ability to specify a key file location, but requiring seems a little excessive to me since with the private key in `~/.ssh/id_rsa`, `git checkout git@...` will work without issues?

sonicaj:
I think we should add that path as a default and if no override is specified and neither is a path specified in the manifest, it would use the default and if the path does not exist, error out in that case.

yocalebo:
> I think we should add that path as a default and if no override is specified and neither is a path specified in the manifest, it would use the default and if the path does not exist, error out in that case.

Yep, I agree

sonicaj:
Done :)

scale_build/packages/git.py
```diff
@@ -117,6 +123,43 @@ def checkout(self, branch_override=None, retries=3):
...
+        if self.ssh_based_source:
+            return [
+                'git', '-c',
+                f'core.sshCommand=ssh -i {self.get_identity_file_path} -o StrictHostKeyChecking=\'accept-new\''
+            ]
+        else:
+            return ['git']
+
+    @property
+    def get_identity_file_path(self):
```
yocalebo, line 137: We should enforce that an absolute path is given to us. `os.path.abspath` doesn't work, for example, if user sets something like `~/.ssh/id_rsa` (I just hit this problem testing this branch 😄 )

sonicaj: This should cater to that https://github.com/truenas/scale-build/pull/323/commits/bdc665586a4418053090852db1fd6d630144afeb. Let me know if you spot any issues on that end please 🤔 

I just think it is convenient to be able to put any custom keys in `identities/` and then specify an env variable `./identities/abckey` from CI perspective as this way the builder can live anywhere on the host

conf/build.manifest
```diff
@@ -1,6 +1,7 @@
 ---
 code_name: "Bluefin"
 debian_release: "bullseye"
+identity_file_path_default: "~/.ssh/id_rsa"
```
yocalebo, line 4: We'll have to expand the current users home directory or this won't work. i.e. `os.path.expanduser('~') + /.ssh/id_rsa`

yocalebo: Oh I see you're doing this down below.

yocalebo (approved):
Built an image and verified corssl was installed and functional.

PR #333, opened 2022-08-21
NAS-117779 / 22.12 / Enforce passwd/group specified reference files

sonicaj:
This PR adds changes to enforce passwd/group reference files in the basechroot to prevent changes of uid/gid in installations.
The problem we faced was that some users/groups might not get installed in subsequent builds resulting in changing gids as in few different packages, the users/groups they install is not based on constant uid/gid but rather generated incrementally based on other packages.

What we do here is split the debootstrap process in 2 phases where in the first phase we just download relevant packages without installing and then before the packages are configured/installed ( where actually all users/groups are created ), we enforce the specified users/groups and then continue with the debootstrap process.


themylogin (approved):
Awesome, thank you for taking care of this!

scale_build/bootstrap/bootstrapdir.py
```diff
@@ -145,6 +149,22 @@ def after_extra_packages_installation_steps(self):
...
+    def debootstrap_debian(self):
+        manifest = get_manifest()
+        run(
+            ['debootstrap'] + self.deopts + [
+                '--foreign', '--keyring', '/etc/apt/trusted.gpg.d/debian-archive-truenas-automatic.gpg',
+                manifest['debian_release'],
+                self.chroot_basedir, manifest['apt-repos']['url']
+            ]
+        )
+        for reference_file in REFERENCE_FILES:
```
themylogin, line 161: Actually... Do passwd and group exist at this moment? Will you be able to ensure that they don't contain any new lines (i.e. we don't delete any new built-in group or user by overwriting files)

sonicaj: They don't exist at this moment, so it's clean that way

PR #336, opened 2022-08-26
NAS-117759 / 22.12 / Set 0o700 permissions for data directory

Qubad786:
## Context

It was requested that we set `600` permissions for database file to avoid it being world readable.

### DIFF
--- a/truenas_install/__main__.py
+++ b/truenas_install/__main__.py
@@ -363,6 +363,10 @@ def main():
 
                     setup_machine_id = configure_serial = True
 
+                # We do not want /data directory to be world readable
+                # Doing this here is important so that we cover both fresh install and upgrade case
+                run_command(["chmod", "-R", "u=rwX,g=,o=", f"{root}/data"])
+
                 if setup_machine_id:
                     with contextlib.suppress(FileNotFoundError):
                         os.unlink(f"{root}/etc/machine-id")


anodos325:
I believe original ticket was to change /data to 700? Was there some reason this wasn't possible? If so, perhaps a dev-facing note in the ticket is warranted.

truenas_install/__main__.py
```diff
@@ -363,6 +363,8 @@ def main():
 
                     setup_machine_id = configure_serial = True
 
+                run_command(["chmod", "700", f"{root}/data"])
```
william-gr, line 366: Seems you missed -R

anodos325: Or perhaps `chmod -R u=rwX,g=,o= <path>`
Which will set  0700 on dirs and 0600 on files.

themylogin: @anodos325 just to confirm will that syntax work on FreeBSD?

anodos325:
```
root@homenas[/mnt/dozer]# chmod -R u=rwX,g=,o= f
root@homenas[/mnt/dozer]# stat -x f/testfile    
  File: "f/testfile"
  Size: 0            FileType: Regular File
  Mode: (0600/-rw-------)         Uid: (    0/    root)  Gid: (    0/   wheel)
Device: 18446744072137047106,1457848327   Inode: 426380    Links: 1
Access: Mon Aug 29 13:36:04 2022
Modify: Mon Aug 29 13:36:04 2022
Change: Mon Aug 29 13:36:54 2022
root@homenas[/mnt/dozer]# cat /etc/version 
TrueNAS-12.0-U8.1 (4e9efc861d)
```

Qubad786: Do we want to recursively set permissions as 700 for all files?
This will set executable bit for database as well..

anodos325: > Do we want to recursively set permissions as 700 for all files? This will set executable bit for database as well..

See my chmod incantation above. It sets 700 for dirs and 600 for files. `X` vs `x`.

yocalebo (commented):
Let's add a comment on what this is doing exactly so anyone that comes after understands 😄

yocalebo (approved):

PR #338, opened 2022-09-06
NAS-117794 / 22.12 / Remove builder resolvconf from SCALE iso

sonicaj:


yocalebo:
While this method seems to work, it's not the "proper" way when creating a squashfs filesystem. Proper way is to either
1. create a file and write these entries on a newline and then specify the `-ef <path-to-exclude-list-file>` flag when calling `mksquashfs` command
2. or specifying the dir/file to be excluded via `-e <dir-1> <dir-2> <file-1>` etc etc

yocalebo (approved):

PR #343, opened 2022-09-19
NAS-111127 / 22.12 / Load drivetemp module

SuperQ:
Load the `drivetemp` module[0] in order to allow monitoring of disk temps via hwmon/lm-sensors.

[0]: https://www.kernel.org/doc/html/latest/hwmon/drivetemp.html

Signed-off-by: SuperQ <superq@gmail.com>

### DIFF
--- a/conf/cd-files/etc/modules-load.d/drivetemp.conf
+++ b/conf/cd-files/etc/modules-load.d/drivetemp.conf
@@ -1,0 +1,3 @@
+# TrueNAS managed
+# Load drivetemp to allow monitoring via hwmon.
+drivetemp


william-gr:
Thank you for your contribution!

yocalebo:
First, thanks for the contribution! Unfortunately, this kernel module only reports ATA devices (no SAS devices) and devices that are directly connected (i.e. doesn't report disk temps in JBODs or external enclosures). I'll have to investigate a bit more on how this can be integrated into our product.

SuperQ:
Yea, it also doesn't help with USB drives.

But, some data is better than no data.

Really, I would love to start adding some Prometheus exporters, for example, [smartclt_exporter](https://github.com/prometheus-community/smartctl_exporter) could be included to provide a lot more / better data.

But for now, I'm planning to deploy this things as Kubernetes pods.

yocalebo:
@SuperQ [smartclt_exporter ](https://github.com/prometheus-community/smartctl_exporter) looks like all it's doing is wrapping around the `smartctl` command and fork+exec'ing out to the CLI to gather that information. That's actually a pretty bad way of getting that information unfortunately. We can't put that in our product because it doesn't scale on systems with many disks. We've learned this the hard way. 😄

SuperQ:
Oh yes, I'm very aware of the dangers and problems with fork/exec. It's explicitly banned from most of the monitoring collectors I maintain.

Unfortunately, on Linux, I have not found any other standard interface for SMART data. This is why it's separated out into a stand-alone tool, and not part of something like the [node_exporter](https://github.com/prometheus/node_exporter).

I've been considering https://github.com/anatol/smart.go as a possible replacement for the exec'ing that the exporter is currently doing.

yocalebo:
I have investigated that golang library extensively and can confirm, that's the "proper" solution BY FAR. It's a shame that nothing exists for python. However, I have a local branch where I've been slowly writing a pure python implementation that gets smart related data using `ioctl` 's so no fork+exec'ing involved. It's slow going and insanely involved. The problem, is that there are 1000's of types of drives and understanding each of their quirks is impossible at this point. That's why `smartmontools` is the leader in this respect since they've been around for so long and actually keep a "database" of "quirks" for various models of drives.

That golang project actually downloads the smartmontools header file, dynamically writes a C source file that parses the header file and spits out "formatted" data to stdout, compiles their C file and then runs it to generate a golang hash map. That's how they're working around that particular problem.

Anyways, something to keep in mind. `smartd` actually writes various smart related attributes to files for each disk. You can read the files on disk and get the data as necessary. As a matter of fact, I have a remote branch in `truenas/middleware` repo called `optimize-smart-temp` that goes and reads these files to get the drive temperatures. The only problem is that `smartmontools` hasn't added functionality to write temperature data for nvme devices to a file.....however, linux kernel seems to report that by default (if the nvme drive supports it of course)

freqlabs:
This is just loading it in the installer isn't it?

yocalebo:
@SuperQ thanks for the commit but we're not going to do this. We already do keep track of drive temperatures and that information can be accessed using our API. This won't add any immediate benefit to us.

SuperQ:
Sounds fine. Maybe if we added a Prometheus endpoint that exposed this data, that would be a better long-term solution anyway.

PR #344, opened 2022-09-19
[NAS-118248] Add sysfsutils package

SuperQ:
Add the `sysfsutils` package to the build manifest. This allows simple tweaks of file attributes in the `/sys` filesystem.

Signed-off-by: SuperQ <superq@gmail.com>

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -237,6 +237,8 @@ additional-packages:
   comment: requested by community (NAS-108490)
 - package: ipmctl
   comment: requested by community (NAS-108490)
+- package: sysfsutils
+  comment: requested by community (NAS-118248)
 
 #
 # List of additional packages installed into TrueNAS SCALE ISO file


freqlabs:
I'm not convinced this is that useful. It's already pretty simple to manipulate files in sysfs.

SuperQ:
It's simple, but this package makes it easy to make it persistent across reboots. That's the "hard" part that this helps solve.

yocalebo:
It also allows them to be persistent across reboots and upgrades.

SuperQ:
@yocalebo Nice. Will that also allow for sysfs filesystem permission/ownership changes?

SuperQ:
Note, my plan was to also add API support to control the tuning features that this package allows for. This was just to get the idea rolling and provide the framework (`/etc/sysfs.d/*.conf`) to do those configurations.

SuperQ:
My use case is to adjust the permissions/group (for group read) of `/sys/class/powercap/intel-rapl:*/energy_uj` to allow monitoring software to collect metrics as non-root. Otherwise the monitoring agent needs root, which I like to avoid.

freqlabs:
> It's simple, but this package makes it easy to make it persistent across reboots. That's the "hard" part that this helps solve.

This can be done easily enough with [tmpfiles.d](https://www.freedesktop.org/software/systemd/man/tmpfiles.d.html) already. I find it best to avoid relying on additional dependencies especially during system startup.

SuperQ:
Oh, that's an interesting option. I'll check it out to see if it works for this.

yocalebo:
As discussed earlier, I've got a branch that I haven't pushed that adds the ability to change sysfs settings via our API so there is no need for adding this at this time. I'll consider adding the "perms" portion as I work on it more.

SuperQ:
That sound great, thanks!

PR #346, opened 2022-10-05
NAS-118432 / 22.12 / add acpidump tool

yocalebo:


william-gr (approved):

themylogin (commented):
We have install-dev-tools script, why don’t we add things like this there? The image is large enough already.

PR #362, opened 2022-11-29
Change code name and version to Cobia

william-gr:


conf/build.manifest
```diff
@@ -1,5 +1,5 @@
 ---
-code_name: "Bluefin"
+code_name: "Cobia"
```
william-gr, line 2: @sonicaj I am not sure where this is used, tbh

sonicaj (approved):

conf/build.manifest
sonicaj: This is used in the manifest we add to update image

PR #364, opened 2022-12-02
Update Bluefin nightly Artifacts job name in Publish ISO and UPdate

ericbsd:


william-gr (commented):
Which branches is this using?

ericbsd:
> Which branches is this using?

Should be stable/bluefin

william-gr (approved):

PR #372, opened 2023-01-06
NAS-119473 / 23.10 / Use patched parted

themylogin:


yocalebo (approved):
Since Jenkins custom builder is down, I tested this on my custom builder and it built without issues. Approving

themylogin:
time 6:00

PR #377, opened 2023-02-24
Build production and debug kernels for TrueNAS SCALE

usaleem-ix:
This commits adds support for production and debug kernels. Debian packages for production and debug kernels are built separately. Kernel headers and images for both production and debug kernels will be installed on the base system.

Since there are two kernels images that would be present on the system, kernel modules should also be built for both kernels. Debian packages for OpenZFS and SCST kernel modules are built separately for production and debug kernels.

OpenZFS and SCST kernel modules for debug kernel would have debug flags enabled. The debug kernel has KGDB enabled to debug the system during boot and runtime.

This PR should be merged after all of the following have been merged:

- https://github.com/truenas/linux/pull/69
- https://github.com/truenas/middleware/pull/10741
- https://github.com/truenas/scst/pull/6

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -57,9 +57,11 @@ apt-repos:
 ############################################################################
 base-packages:
 - dosfstools
-- linux-truenas-libc-dev
-- linux-headers-truenas-amd64
-- linux-image-truenas-amd64
+- linux-truenas-production-libc-dev
+- linux-headers-truenas-production-amd64
+- linux-headers-truenas-debug-amd64
+- linux-image-truenas-production-amd64
+- linux-image-truenas-debug-amd64
 - linux-perf
 - avahi-daemon
 - avahi-utils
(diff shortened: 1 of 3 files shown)


sonicaj (commented):
I am curious, does the final image have all the debug packages installed ?
Not sure if the packages you have added in `base-packages` cover all those debug packages ?

usaleem-ix:
> I am curious, does the final image have all the debug packages installed ? Not sure if the packages you have added in `base-packages` cover all those debug packages ?

All the packages are installed on the system except for the debug info packages for production and debug kernels.

I think we don't need to mention the packages for kernel modules for production kernel in `base-packages` as they are being installed by default, from dependencies of `openzfs` and `scst` packages.

I have added `openzfs-zfs-modules-dbg` and `scst-dbg` in `base-packages`. These packages install kernel modules for OpenZFS and SCST for debug kernel.

I have not added the packages that contain debug symbols for production and debug kernels here, as they are ~650 MB in size each, and it does not make sense to have them installed on base system by default.

yocalebo (approved):
I've confirmed that this works and running through the full CI produces no obvious regressions.

PR #381, opened 2023-03-09
NAS-120648 / 23.10 / Add bison package to iso

anodos325:
bison is a rutime dependency for samba with
elasticsearch spotlight backend.

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -80,6 +80,7 @@ base-packages:
 - grub-efi-amd64-bin
 - htop
 - ifstat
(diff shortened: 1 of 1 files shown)


conf/build.manifest
```diff
@@ -80,6 +80,7 @@ base-packages:
 - grub-efi-amd64-bin
 - htop
 - ifstat
+- bison
```
sonicaj, line 83: If it's samba's dependency, we should be adding it in samba debian package ?

PR #388, opened 2023-03-28
NAS-117638 / 23.10 / Do not clone same repository when single repo hosts multiple packages

sonicaj:
This PR adds changes to make sure we don't clone a single repo multiple times just because it hosts more then 1 package i.e truenas/middleware.


yocalebo:
retest this please

yocalebo (approved):

PR #398, opened 2023-04-14
Do not run lintian to decrease build time

sonicaj:


themylogin (commented):
Should we still use it for production builds?

sonicaj:
It is used to enforce debian package standards and we don't enforce that our debian packages conform to those standards already. 
Essentially what i am trying to say is that linitian is not already being taken seriously and we are just running it as part of the build process and not respecting it's say. So this should be fine if seen from that context.

themylogin (approved):

PR #408, opened 2023-05-02
NAS-121753 / 23.10 / Precheck for S3 service

themylogin:


themylogin:
time 6:00

themylogin:
backport

yocalebo (changes requested):
Some flake8 issues

truenas_install/__main__.py
```diff
@@ -235,17 +246,62 @@ def configure_system_for_zectl(boot_pool):
...
+                                return (
+                                    "You have built-in S3 service enabled in your configuration. This is not supported "
+                                    "anymore. Please, contact our support team."
+                                )
+                        except Exception:
+                            pass
+
+                    for p in psutil.process_iter():
+                        if p.name() == "minio":
+                            # FIXME: ignore if minio is running in k3s
```
yocalebo, line 279: We'll need to solve this before merge. The process is going to be that the customer will start the minio app alongside the host S3 service and add the app as a node to their existing cluster and minio will transfer the data natively.

Then the host service will be shut-off and the app will continue to run so I predict this will fail 100% of the time?

truenas_install/__main__.py
```diff
@@ -235,17 +246,62 @@ def configure_system_for_zectl(boot_pool):
...
+def precheck(old_root):
+    if old_root is not None:
+        if licenseobj := read_license(old_root):
+            if ContractType(licenseobj.contract_type) in [ContractType.silver, ContractType.gold]:
+                if licenseobj.contract_end > datetime.utcnow().date():
+                    db_path = database_path(old_root)
+                    if os.path.exists(db_path):
+                        try:
+                            if query_row(
+                                "SELECT * FROM services_services WHERE srv_service = 's3' AND srv_enable = 1",
```
yocalebo, line 267: Just checking to make sure I fully understand but you've already merged a commit into master that removes this column from this table so this will never be true? But I'm assuming this logic gets executed before the alembic migrations are applied since those get applied during first boot after upgrade?

truenas_install/__main__.py
themylogin: I would be super glad to do it, but apps are broken in Cobia. Can we create a ticket instead?

truenas_install/__main__.py
themylogin: You are correct, the column will be removed on the first boot of the upgraded system. This code is executed way before and can yield both positive and negative results. If the SQL query fails (e.g. we are upgrading 23.10 to 23.10.1 and the columns is removed already) it will be treated as built-in minio service not running because if the column is removed then there is definitely no built-in minio service.

truenas_install/__main__.py
```diff
@@ -235,17 +246,68 @@ def configure_system_for_zectl(boot_pool):
...
+
+                    for p in psutil.process_iter():
+                        if p.name() == "minio":
+                            try:
+                                with open(f"/proc/{p.pid}/cgroup") as f:
+                                    cgroups = f.read()
+                            except FileNotFoundError:
+                                cgroups = ""
+
+                            if "kubepods.slice" not in cgroups:
```
themylogin, line 285: @sonicaj is this a good method to check whether the `minio` process is running as an app?

sonicaj: Yes this should be a good way to identify that

sonicaj (approved):

truenas_install/__main__.py
themylogin: @sonicaj sorry, something was off locally, they are working after a reboot

yocalebo (approved):

PR #418, opened 2023-05-15
NAS-121741 / 23.10 / remove execute bit on dpkg binary

yocalebo:
We already remove the execute bit on the apt* related binaries, but those are simply a wrapper around dpkg. We need to remove execute bit on dpkg since it is the binary that's actually responsible for installing the binary.


yocalebo:
retest this please

bmeagherix (approved):

PR #429, opened 2023-05-24
bump certifi to 2023.05.07

yocalebo:
Certifi 2022.12.07 removes root certificates from "TrustCor" from the root store. These are in the process of being removed from Mozilla's trust store.

TrustCor's root certificates are being removed pursuant to an investigation prompted by media reporting that TrustCor's ownership also operated a business that produced spyware. Conclusions of Mozilla's investigation can be found [here](https://groups.google.com/a/mozilla.org/g/dev-security-policy/c/oxX69KFvsm4/m/yLohoVqtCgAJ).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-certifi==2020.12.5
+certifi==2023.05.07
 chardet==4.0.0
 coloredlogs==15.0
 humanfriendly==9.1


yocalebo:
retest this please

anodos325 (approved):

PR #438, opened 2023-05-31
Add _chrony user and group

bmeagherix:


themylogin (approved):

themylogin:
Donee also need to remove ntp-related users and groups?

bmeagherix:
> Do we also need to remove ntp-related users and groups?

Probably, but when I tried to replace the `ntp` user with `_chrony` the tools complained, so I avoided (for now).  I'll circle back to do that after the middleware changes get merged.

PR #439, opened 2023-06-01
NAS-121623 / 23.10 / Allow not installing recommended packages of a package

sonicaj:
This PR adds changes to allow us to not install recommended packages of some package. Right now we default to assuming that each package wants recommended packages to be installed however we can move ahead with the pruning slowly and have a default where recommended packages are not brought in by default and only in some cases where it is required we can bring those in.

We also fix truecommand here as it is not functional atm because of `wireguard-tools` package missing and if we install its recommended packages it brings in `wireguard-dkms` which we don't want as then it starts pulling upstream linux kernel headers.

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -205,42 +205,61 @@ apt_preferences:
 additional-packages:
 - package: xtail
   comment: used by support (NAS-108788)
+  install_recommends: true
 - package: iperf3
   comment: requested by sales (NAS-108787)
+  install_recommends: true
 - package: fio
   comment: requested by sales (NAS-108787)
+  install_recommends: true
 - package: dnsutils
   comment: requested by community (NAS-109391)
+  install_recommends: true
 - package: traceroute
   comment: requested by platform and perf (NAS-110493)
+  install_recommends: true
 - package: openseachest
   comment: requested by performance team (NAS-106154)
+  install_recommends: true
 - package: cxgbtool
   comment: requested by OS team (NAS-111041)
+  install_recommends: true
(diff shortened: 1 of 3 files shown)


conf/build.manifest
```diff
@@ -85,6 +85,8 @@ base-packages:
 - squashfs-tools
 - sysstat
 - truenas
```
yocalebo, line 87: I think we should just change all of these packages to be prefixed with `name:` to keep it clean and concise. It will also allow for future flexibility if we ever have to add more options for each package.

yocalebo (changes requested):

scale_build/image/update.py
```diff
@@ -79,11 +79,20 @@ def install_rootfs_packages_impl():
...
     ):
-        logger.debug('Installing %r', package)
-        run_in_chroot(['apt', 'install', '-V', '-y', package])
+        if isinstance(package_entry, str):
+            package = package_entry
+            install_recommends = True
+        else:
+            package = package_entry['name']
+            install_recommends = package_entry.get('install_recommends', True)
+
```
yocalebo, line 91: Simplify this logic:
```
log_msg = f'Installing {package_entry['name']!r}'
cmd = ['apt', 'install', '-V', '-y', package_entry['name']]
if not package_entry.get('install_recommends', True):
    log_msg += ' (no recommends)'
    cmd.append('--no-install-recommends')

logger.debug(log_msg)
run_in_chroot(cmd)

yocalebo (approved):
Fantastic! Thank you!

PR #459, opened 2023-07-04
Add socat to allow port-forwarding

jeroenvdheuvel:
Kubernetes port-foward (k3s kubect port-foward) needs `socat` to work. Otherwise an error like this will occur:
> E0704 11:17:00.298021  220736 portforward.go:406] an error occurred
> forwarding 8080 -> 8080: error forwarding port 8080 to pod
> b9936b4a19a6d7a1f1622c9fa5dba657355a494c901a7bade899858fb731a6a6,
> uid : **unable to do port forwarding: socat not found**
>
> E0704 11:17:00.298234  220736 portforward.go:234] lost connection to pod

By installing socat it becomes possible to forward traffic to Kubernetes/k3s services via a command like this:
> k3s kubectl port-forward service/argo-cd-argocd-server -n argo-cd
> 8080:443 --address=0.0.0.0

This change should resolve the bug I raised: https://ixsystems.atlassian.net/browse/NAS-122785

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -273,6 +273,8 @@ additional-packages:
   comment: requested by platform team (NAS-120155)
 - package: i2c-tools
   comment: requested by platform team (NAS-120155)
+- name: socat
+  comment: requested by community (NAS-122785)
 
 #
 # List of additional packages installed into TrueNAS SCALE ISO file


jeroenvdheuvel:
@yocalebo, thanks for looking into it. Can you explain why the change is rejected? I understand that socat might not be necessary for Truenas scale to work. Truenas scale however does include Kubernetes via k3s. The port-forwarding (default functionality of Kubernetes) doesn't work because it's missing socat.

yocalebo:
@jeroenvdheuvel Sorry, I'm not sure what happened to my comment...

1. you're trying to merge into `release/22.12.3.2` but we have already shipped 22.12.3.2 to the public so this PR won't change anything (this isn't something I would expect you to know since this is all related to our internal build process)
2. k3s should ship with all dependent binaries to function so adding socat to the host shouldn't be needed. This is potentially the root of the problem (we'll have to investigate why k3s isn't shipping with this dependency)
3. we're not seeing this error in our latest nightly image (next major TrueNAS SCALE release) so this isolates it to the bluefin release train

We're investigating why k3s in Bluefin isn't shipping with all the necessary dependencies. We'll use your ticket that you opened up here: https://ixsystems.atlassian.net/browse/NAS-122785 to investigate what we find and/or fix 😄

jeroenvdheuvel:
Thanks for letting me know @yocalebo. I was also surprised by the port-forward didn't work. Great that you're looking into it!

PR #464, opened 2023-07-11
Remove build tools from rootfs

themylogin:
A rootfs image was used for package builder (where build tools are needed) and then it was reused for building the NAS rootfs image. Having them separate reduces the image size by about 150 megabytes. TrueNAS does not have gcc installed anymore.


themylogin:
A rootfs image was used for package builder (where build tools are needed) and then it was reused for building the NAS rootfs image. Having them separate reduces the image size by about 150 megabytes

PR #482, opened 2023-08-14
Add systemd-container to additional-packages list

Jip-Hop:
As requested in [NAS-123533](https://ixsystems.atlassian.net/browse/NAS-123533).

The systemd-container package, which provides the systemd-nspawn and machinectl commands, was included with SCALE releases prior to version 22.12.3. It was then gone for some releases and seems to be present again in a future version (tested with TrueNAS-SCALE-23.10-MASTER-20230813-042924).

This pull requests adds systemd-container to [the list of additional-packages](https://github.com/truenas/scale-build/blob/release/23.10-BETA.1/conf/build.manifest#L227) to be installed.

Reasons to include it in the list:

- It was included in previous releases and seems to be included in future releases
- Putting it on the list will ensure the package is explicitly included, instead of being installed as an implicit dependency (and won’t disappear suddenly)
- It’s a very small package (about 1MB installed)
- There’s [a community of users](https://github.com/Jip-Hop/jailmaker) relying on systemd-nspawn on SCALE (and we prefer not to have to resort to using apt to bring back systemd-nspawn)
- No reports of systemd-nspawn breaking Apps

[NAS-123533]: https://ixsystems.atlassian.net/browse/NAS-123533?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -282,6 +282,9 @@ additional-packages:
 - name: i2c-tools
   comment: requested by platform team (NAS-120155)
   install_recommends: true
+- name: systemd-container
+  comment: requested by community (NAS-123533)
+  install_recommends: true
 
 #
 # List of additional packages installed into TrueNAS SCALE ISO file


yocalebo:
@Jip-Hop thanks a lot for the PR and the very detailed justification for wanting to add this to our base image. It allowed me to immediately understand all the reasons for why this change is being requested so I greatly appreciate it.

Unfortunately, however, we're in the perpetual process of removing 3rd party packages/dependencies that our api does not actively use. We're really trying to pay off the proverbial "CVE technical debt" by shrinking our attack surface. Because of this fact, we will not be accepting this change.

As an aside, I've found that you've written a _VERY_ neat and interesting "jail" script that is run on SCALE that very clearly makes alterations to the base image that is not supported by us. I believe it's only natural to update your script to check for the existence of the systemd-container package and add it if it's not already included. This would at least allow that portion of the community to be able to continue using your solution 😄 

NOTE: (I'm sure you're aware of how to do this already) But to add a package to scale, you'll need to `chmod +x` the apt binaries as well as the `dpkg` binary. Once that's complete, you can update the local cache and then install the package as necessary. Understanding, of course, that the changes will be lost on upgrade.

Jip-Hop:
Thanks for the compliment on the jail script! 😄 In my opinion it was more neat when it was 'living off the land' using the packages provided by default and interfering as little as possible. But I had already implemented the `chmod +x /bin/dpkg* /bin/apt*` method to install `systemd-container` when it was missing.

Are there any plans to progressively prevent adding a package to SCALE (like removing the package manager completely)? Since first it was just apt `/bin/apt*` then later also `/bin/dpkg*`.

Do you have an idea why systemd-container was gone in the `22.12.3` releases? I found that the reason it was installed before and in `23.10` is because the `truenas` package depends on `libvirt-daemon-system` which depends on `libvirt-daemon-system-systemd` which finally installs `systemd-container`.

```shell
# TrueNAS-SCALE-23.10-MASTER-20230813-042924
# With systemd-container

admin@truenas[~]$ apt list '?any-version(?installed?depends(?exact-name(systemd-container)))'
Listing... Done
libnss-mymachines/bookworm,now 252.6-1 amd64 [installed,automatic]
libvirt-daemon-system-systemd/bookworm,now 9.0.0-4 all [installed,automatic]
admin@truenas[~]$ apt list '?any-version(?installed?depends(?exact-name(libnss-mymachines)))' 
Listing... Done
admin@truenas[~]$ apt list '?any-version(?installed?depends(?exact-name(libvirt-daemon-system-systemd)))'
Listing... Done
libvirt-daemon-system/bookworm,now 9.0.0-4 amd64 [installed,automatic]
admin@truenas[~]$ apt list '?any-version(?installed?depends(?exact-name(libvirt-daemon-system)))'        
Listing... Done
truenas/now 20230813053755~truenas+1 all [installed,local]
```

But in `22.12.3` (without `systemd-container`) the `truenas` package still depends on `libvirt-daemon-system` but instead of the `libvirt-daemon-system-systemd` dependency, it went with an alternative package instead: `libvirt-daemon-system-sysv`. And because `libvirt-daemon-system-sysv` doesn't depend on `systemd-container` that seems to be the reason why it was gone.

According to the description of [libvirt-daemon-system-systemd](https://packages.debian.org/bullseye/libvirt-daemon-):

> This package contains the dependencies to make libvirt work with systemd. **(this is the default)**.

And the description of [libvirt-daemon-system-sysv](https://packages.debian.org/bullseye/libvirt-daemon-system-sysv):

> This package contains dependencies and init scripts to make libvirt work with sysv based init systems. **Using libvirt-daemon-systemd is preferred** since the init scripts are incomplete.

So somehow `22.12.3` started to deviate from what is recommended and the default. But I can't understand why. Do you have an idea?

PR #494, opened 2023-09-11
NAS-124055 / 24.04 / Implement dataset-based filesystem hierarchy

anodos325:
Create filesystem hierarchy for new boot environments based
on specifications in fhs.py file in truenas_install directory.

This gives us more flexibility regarding which parts of FS to
make readonly, and also identifying precise local changes from
a pristine environment that users have made to system files.

Precise settings detailed in comments in the specification file.
Overall, this gives better posture for STIG compliance regarding
auditability and prevention of unauthorized OS changes.

### DIFF
--- a/truenas_install/__main__.py
+++ b/truenas_install/__main__.py
@@ -423,13 +423,13 @@ def main():
             for entry in TRUENAS_DATASETS:
                 this_ds = entry['name']
                 ds_name = f"{dataset_name}/{this_ds}"
-                ds_path = entry.get("mp") or entry["name"]
-                ds_guid = run_command(["zfs", "list", "-o", "guid", "-H", ds_name])
+                ds_path = entry.get("mountpoint") or f"/{entry['name']}"
+                ds_guid = run_command(["zfs", "list", "-o", "guid", "-H", ds_name]).stdout.strip()
 
-                mp = os.path.join(root, ds_path[1:] if ds_path.startswith('/') else ds_path)
+                mp = os.path.join(root, ds_path[1:])
                 os.makedirs(mp, exist_ok=True)
                 run_command(["mount", "-t", "zfs", f"{dataset_name}/{this_ds}", mp])
-                ds_info.append({"ds": ds_name, "guid": ds_guid.stdout.strip(), "fhs_entry": entry})
+                ds_info.append({"ds": ds_name, "guid": ds_guid, "fhs_entry": entry})
                 undo.append(["umount", mp])
 
             try:
@@ -472,13 +472,7 @@ def main():
 
                     os.chmod(f"{root}/{entry['name']}", force_mode)
 
-                # /usr will be readonly and so we want the ca-certificates directory to symlink
-                # to writeable location in /var/local
-                os.makedirs(f"{root}/usr/local/share", exist_ok=True)
-                shutil.rmtree(f"{root}/usr/local/share/ca-certificates", ignore_errors=True)
-                os.symlink("/var/local/ca-certificates", f"{root}/usr/local/share/ca-certificates")
-
-                with open(f"{root}/conf/truenas_root_ds.conf", "w") as f:
+                with open(f"{root}/conf/truenas_root_ds.json", "w") as f:
                     f.write(json.dumps(ds_info, indent=4))
 
                 with contextlib.suppress(FileNotFoundError):
@@ -698,7 +692,7 @@ def main():
 
         for entry in TRUENAS_DATASETS:
             this_ds = f"{dataset_name}/{entry['name']}"
-            mp = entry.get('mp') or f"/{entry['name']}"
+            mp = entry.get('mountpoint') or f"/{entry['name']}"
             ro = "on" if "RO" in entry["options"] else "off"
             run_command(["zfs", "set", f"readonly={ro}", this_ds])
 
--- a/truenas_install/__main__.py
+++ b/truenas_install/__main__.py
@@ -18,6 +18,7 @@ import textwrap
 import psutil
 
 from licenselib.license import ContractType, License
+from .fhs import TRUENAS_DATASETS
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


anodos325:
This PR depends on https://github.com/truenas/middleware/pull/12075 (WIP)

truenas_install/fhs.py
```diff
@@ -0,0 +1,105 @@
...
+    },
+    {
+        'name':  'var',
+        'options': ['NOSUID', 'NOACL', 'NOATIME'],
+        'snap': True
+    },
+    {
+        'name':  'var/ca-certificates',
+        'options': ['NOSUID', 'NOACL', 'NOEXEC'],
+        'mp': '/var/local/ca-certificates'
```
sonicaj, line 99: Should we maybe use `mountpoint` instead ?

truenas_install/__main__.py
```diff
@@ -434,6 +466,21 @@ def main():
 
                 write_progress(0.5, "Performing post-install tasks")
 
+                for entry in TRUENAS_DATASETS:
+                    if not (force_mode := entry.get("mode")):
+                        continue
+
+                    os.chmod(f"{root}/{entry['name']}", force_mode)
```
sonicaj, line 473: Do we want this to be recursive or the option to be configurable ?

truenas_install/fhs.py
```diff
@@ -0,0 +1,105 @@
...
+opt - see FHS
+usr - see FHS
+var - see FHS
+var/log - separate dataset for system logs. This is to provide flexibility to
+    snapshot and replicate log information as needed by administrator.
+var/ca-certificates - administrator-provided CA certificates, symlinked
+    from /usr/local/share/ca-certificates.
+"""
+
+TRUENAS_DATASETS = [
```
sonicaj, line 50: We can probably use json schema validation here to ensure that this is valid and the github CI action can catch that (can be done separately too in a ticket i guess :P)

truenas_install/__main__.py
```diff
@@ -434,6 +466,21 @@ def main():
...
+
+                # /usr will be readonly and so we want the ca-certificates directory to symlink
+                # to writeable location in /var/local
+                os.makedirs(f"{root}/usr/local/share", exist_ok=True)
+                shutil.rmtree(f"{root}/usr/local/share/ca-certificates", ignore_errors=True)
+                os.symlink("/var/local/ca-certificates", f"{root}/usr/local/share/ca-certificates")
+
+                with open(f"{root}/conf/truenas_root_ds.conf", "w") as f:
```
sonicaj, line 481: Should we maybe have it named as `.json` instead ?

truenas_install/fhs.py
anodos325: Okay. Makes sense.

truenas_install/__main__.py
anodos325: Since it's a newly created dataset it won't have contents (unless squashfs already had it - in which case it should probably be fixed in package install).

truenas_install/__main__.py
anodos325: Done

truenas_install/fhs.py
themylogin: Also we can set default variable values based on the same schema

truenas_install/__main__.py
```diff
@@ -389,16 +390,48 @@ def main():
...
     try:
         with tempfile.TemporaryDirectory() as root:
+            undo = []
+            ds_info = []
             run_command(["mount", "-t", "zfs", dataset_name, root])
+            for entry in TRUENAS_DATASETS:
+                this_ds = entry['name']
+                ds_name = f"{dataset_name}/{this_ds}"
+                ds_path = entry.get("mountpoint") or entry["name"]
+                ds_guid = run_command(["zfs", "list", "-o", "guid", "-H", ds_name])
```
themylogin, line 427: It's better to do `.stdout.strip()` so `ds_guid` is really a GUID and not some subprocess data structure

truenas_install/__main__.py
```diff
@@ -389,16 +390,48 @@ def main():
...
+
+                mp = os.path.join(root, ds_path[1:] if ds_path.startswith('/') else ds_path)
```
themylogin, line 429: Just `ds_path.rstrip("/")`, but why are absolute and relative paths are mixed here?

truenas_install/__main__.py
```diff
@@ -434,6 +466,21 @@ def main():
...
+                for entry in TRUENAS_DATASETS:
+                    if not (force_mode := entry.get("mode")):
+                        continue
+
+                    os.chmod(f"{root}/{entry['name']}", force_mode)
+
+                # /usr will be readonly and so we want the ca-certificates directory to symlink
+                # to writeable location in /var/local
+                os.makedirs(f"{root}/usr/local/share", exist_ok=True)
+                shutil.rmtree(f"{root}/usr/local/share/ca-certificates", ignore_errors=True)
```
themylogin, line 478: Shouldn't this be done in rootfs image preparation stage instead?

themylogin (changes requested):

truenas_install/fhs.py
yocalebo: No I don't agree, we are actively trying to remove dependencies on 3rd party tools and I definitely don't want to add this dependency to our installer. It's overkill for what we want to do. We can revisit this request in another ticket but for now, I don't want to do this.

truenas_install/__main__.py
anodos325: I was doing it later, but you're right it makes more sense to do that here.

truenas_install/fhs.py
sonicaj: @yocalebo json schema is already a dependency of scale-build

truenas_install/__main__.py
anodos325: Yeah, that's a better place to put it.

truenas_install/fhs.py
anodos325: It's not a dependency in our installer though.

anodos325: Okay. I created a schema and plumbed into our scale-build validate methods (rather than doing at runtime).

sonicaj (approved):

themylogin (approved):

yocalebo (approved):

PR #496, opened 2023-09-25
NAS-124338 / 24.04 / Make rootfs readonly

anodos325:
This PR expands existing fhs setup to make `/`  and `/opt` readonly, and make `/root` a separate dataset.


yocalebo (approved):
What's the worst that can happen 😉

sonicaj (approved):

PR #546, opened 2023-12-27
NAS-125472 / 24.04 / Validate cached bootstrap directory based off on the reference group/passwd files

sonicaj:
In the SCALE build process, when building the cache, it considers necessary cache files. If these files already exist, the SCALE build doesn't rebuild the cache, assuming the previous cache is sufficient. However, if the user or group information in the reference file has changed, the cache is declared as healthy, and the SCALE build doesn't rebuild the cache.

This PR effectively tackles that problems and ensures the cached bootstrapdir actually reflects the reference group/passwd files and if reference files have been changed, it invalidates the cache and initiates a clean build. Earlier what would happen is that during update phase usually the image won't build as we validate at that point if the users currently existing in the chroot env actually reference the reference group/passwd files.

### DIFF
--- a/scale_build/bootstrap/cache.py
+++ b/scale_build/bootstrap/cache.py
@@ -66,6 +66,9 @@ class CacheMixin:
             else:
                 intact = False
 
+            # Remove the temporary restored cached directory
+            shutil.rmtree(self.chroot_basedir, ignore_errors=True)
+
         if not intact:
             self.remove_cache()
 
--- a/scale_build/bootstrap/cache.py
+++ b/scale_build/bootstrap/cache.py
@@ -1,6 +1,8 @@
 import os
+import shutil
 
 from scale_build.utils.paths import CACHE_DIR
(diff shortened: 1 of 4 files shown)


scale_build/bootstrap/cache.py
```diff
@@ -52,6 +54,21 @@ def mirror_cache_intact(self):
             self.logger.debug('Upstream repo changed! Removing squashfs cache to re-create.')
             intact = False
 
+        if intact:
+            self.restore_cache(self.chroot_basedir)
+            if os.path.exists(
```
themylogin, line 59: I don’t like that checked file names are hardcoded here. Can we please move them away to the same module that compares them?

Also what if some files exist and others don’t? Shouldn’t that be an impossible situation (that causes an exception)?

themylogin (changes requested):

themylogin (approved):
Thank you!

PR #551, opened 2024-01-05
NAS-126052 / 24.04 / Add a debug statement to know if cached squashfs needs to be re-created

sonicaj:


scale_build/bootstrap/cache.py
```diff
@@ -56,9 +56,13 @@ def mirror_cache_intact(self):
 
         if intact:
             self.restore_cache(self.chroot_basedir)
-            for _, diff in compare_reference_files(cut_nonexistent_user_group_membership=True):
+            for reference_file, diff in compare_reference_files(cut_nonexistent_user_group_membership=True):
                 if diff:
                     intact = False
+                    self.logger.debug(
```
themylogin, line 62: Wouldn't it better to present a readable multilines diff rather than `%r`?

themylogin (approved):

PR #565, opened 2024-01-19
TE-1934 Electric Eel updates

alexandrabain:
Updated to master for electric eeel.

### DIFF
--- a/jenkins/FullBuild
+++ b/jenkins/FullBuild
@@ -42,8 +42,8 @@ pipeline {
     }
     stage('Upload') {
       steps {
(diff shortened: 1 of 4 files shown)


kmoore134 (approved):

kmoore134:
@ericbsd Approved now. Need to confirm what update train / tag we build nightly images with. Needs to match where we are tossing the update files online.

kmoore134:
Actually, looks like this is already set properly in the build repo: 

https://github.com/truenas/scale-build/blob/master/conf/build.manifest#L2

ericbsd (approved):
LGTM

PR #572, opened 2024-01-31
NAS-127113 / 24.04-BETA.1 / Explicitly add systemd-container to additional-packages list

Jip-Hop:
As planned in https://ixsystems.atlassian.net/browse/NAS-125733. Ticket is already closed even though systemd-container is not added explicitly as a package to install. It's currently only included in SCALE as [a transient dependency of libvirt-daemon-system](https://github.com/truenas/middleware/blob/release/24.04-BETA.1/debian/debian/control). See my previous attempt at this PR for more info: https://github.com/truenas/scale-build/pull/482.

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -291,6 +291,9 @@ additional-packages:
 - name: i2c-tools
   comment: requested by platform team (NAS-120155)
   install_recommends: true
+- name: systemd-container
+  comment: requested by community (NAS-123533)
+  install_recommends: false
 
 #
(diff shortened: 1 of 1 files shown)


yocalebo:
@Jip-Hop why do we need `nss-mymachines` (this gets brought in via the `install_recommends: True` setting. Do you make use of that in your script?

yocalebo (changes requested):
Oh, this won't make it into BETA.1. We missed that boat already, you'll have to retarget this to the `stable/dragonfish` repo.

PR #573, opened 2024-01-31
Explicitly add systemd-container to additional-packages list

Jip-Hop:
As planned in https://ixsystems.atlassian.net/browse/NAS-125733. Ticket is already closed even though systemd-container is not added explicitly as a package to install. It is currently only included in SCALE as [a transient dependency of libvirt-daemon-system](https://github.com/truenas/middleware/blob/release/24.04-BETA.1/debian/debian/control). See my previous attempt at this PR for more info: https://github.com/truenas/scale-build/pull/482.

Re-targeted https://github.com/truenas/scale-build/pull/572 for the `master` branch as instructed by @kmoore134.

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -291,6 +291,9 @@ additional-packages:
 - name: i2c-tools
   comment: requested by platform team (NAS-120155)
   install_recommends: true
+- name: systemd-container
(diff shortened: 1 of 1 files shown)


kmoore134 (approved):
LGTM

PR #591, opened 2024-02-14
Disallow upgrading from TrueNAS CORE

themylogin:


themylogin:
Replaced by https://github.com/truenas/scale-build/pull/605

PR #592, opened 2024-02-14
Create netdata log directory on update/fresh installation

Qubad786:
## Problem
The Netdata log directory is currently created by the `netdata_setup` function in the middleware. However, the issue is that it runs on every netdata restart, whereas it should only be created once.

## Solution
The Netdata log directory will be created only once during the setup of the new BE (this applied to both fresh installation and update case)

### DIFF
--- a/truenas_install/__main__.py
+++ b/truenas_install/__main__.py
@@ -589,6 +589,10 @@ def main():
 
                     setup_machine_id = configure_serial = True
 
+                # Making log dir for netdata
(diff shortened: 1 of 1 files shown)


themylogin (changes requested):
We should not be doing this in the installer. Neither scatter hardcoded paths and uids/gids all over our code base.

Qubad786:
After discussion with @themylogin we will move this logic to system dataset setup when we setup netdata dataset ✅

PR #607, opened 2024-03-28
NAS-128069 / 24.10 / Update python-netsnmpagent library

anodos325:
This syncs up version of python-netsnmpagent with what is used in TrueNAS core, which includes many fixes (including use-after-free).


yocalebo (approved):

yocalebo:
This should be backported to 24.04.0

anodos325:
stable/dragonfish was addressed by replacing the branch.

PR #619, opened 2024-04-16
NAS-127428 / 24.10 / Bring up all network interfaces on CD boot

themylogin:


conf/cd-files/etc/systemd/network/99-wildcard.network
```diff
@@ -0,0 +1,5 @@
+[Match]
+Name=en*
```
yocalebo, line 2: You'll need to do `Name=*`

themylogin:
time 2:00

sonicaj (approved):

PR #623, opened 2024-04-24
NAS-128497 / 24.10 / remove unmaintained chelsiouwire

yocalebo:
This is unmaintained and I've confirmed that it's shipping a broken `chelsio_adapter_config.py` script which is, essentially, the entire purpose of this repo. This has been broken since SCALE has been released publicly which is why I can justify removing it from the build. I will work with platform team on a proper solution.

NOTE: the only tool we install from this repo is `cxgbtool`. Again, we don't have that on CORE and we're not using it on SCALE. This repo is too large with too much cruft to justify shipping SCALE with a tool that's not even used and shipping a script that doesn't even work.

P.S. linux has proper in-tree drivers for all the various Chelsio cards so we're losing no functionality (with the exception of `cxgbtool` (that we don't even use))

P.S.S. Making a correction. We do, in fact, have cxgbetool on CORE but it's not immediately obvious if we actually use it. Again, the entire point of this repo is to have the ability to flash certain cards to operate a different way. We don't need the cxgbtool (AFAICS) to do this.

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -338,9 +338,6 @@ sources:
   predepscmd:
     - "apt install -y wget xz-utils"
     - "./pull.sh"
-- name: chelsio_uwire
-  repo: https://github.com/truenas/chelsiouwire
-  branch: master
 - name: openseachest
   branch: master
   repo: https://github.com/truenas/openseachest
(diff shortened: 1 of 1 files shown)


themylogin (approved):

sonicaj (commented):
We probably need to remove `cxgbtool` from the manifest as well

sonicaj (approved):

PR #626, opened 2024-04-26
NAS-128564 / 24.10 / Add truenas/sssd to SCALE build

anodos325:
This adds our sssd repository to the build.

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -116,6 +116,8 @@ base-packages:
   install_recommends: true
 - name: truenas-samba
   install_recommends: true
+- name: truenas-sssd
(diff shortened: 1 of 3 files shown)


conf/build.manifest
```diff
@@ -195,6 +197,9 @@ apt_preferences:
 - Package: "*cuda*"
   Pin: "version 525.89*"
   Pin-Priority: 1000
+- Package: "*curl*"
```
yocalebo, line 200: This seems unrelated?

anodos325: runtime requirement of newer curl IIRC.

PR #644, opened 2024-05-21
NAS-129131 / 24.10 / Remove TXT record from mDNS for truenas_installer

anodos325:
We decided against having mDNS update installer status since we won't be consuming the feature.


yocalebo (approved):

yocalebo:
Do we even need mDNS at all?

anodos325:
> Do we even need mDNS at all?

Yes

PR #648, opened 2024-05-29
NAS-129248 / 24.10 / Use pre-built nodejs for truenas_webui

anodos325:
This commit changes the truenas_webui build process so that we download the version specified in the fetch script from upstream, validate the shasum for tarball, and install in the chroot environment for the truenas_webui package build.

This ensures that the webui team can specify which nodejs version to build against without having to rely on upstream debian package.


yocalebo (approved):

anodos325:
jenkins retest this please

PR #653, opened 2024-06-04
NAS-129389 / 24.10 / Add apps validation to build

sonicaj:
This PR removes packages which are no longer consumed because of removal of kubernetes and adds the new validation framework instead in place.


yocalebo (approved):
SHIP IT :smile:

sonicaj:
Just going to run API tests and see where we are with it, we should be good generally speaking ;)

PR #661, opened 2024-07-09
NAS-129797 - adding vendor to build

aiden3c:
If the environment variable `TRUENAS_VENDOR` is passed, upon ISO creation we create a `/data/.vendor` file that is JSON encoded, containing a "name" key with the value passed from TRUENAS_VENDOR.


scale_build/image/iso.py
```diff
@@ -111,7 +119,7 @@ def make_iso_file():
         shutil.copy(os.path.join(CHROOT_BASEDIR, 'usr/share/grub/unicode.pf2'),
                     os.path.join(CD_DIR, 'EFI/debian/fonts/unicode.pf2'))
 
-        iso = os.path.join(RELEASE_DIR, f'TrueNAS-SCALE-{get_image_version()}.iso')
+        iso = os.path.join(RELEASE_DIR, f'TrueNAS-SCALE-{get_image_version(vendor=TRUENAS_VENDOR)}.iso')
```
themylogin, line 122: Why do we need to pass this as a keyword arg?

PR #662, opened 2024-07-09
NAS-129797 / 24.10 / Add .vendor to ISO

aiden3c:
If the environment variable TRUENAS_VENDOR is passed, upon ISO creation we create a /data/.vendor file that is JSON encoded, containing a "name" key with the value passed from TRUENAS_VENDOR.

### DIFF
--- a/scale_build/config.py
+++ b/scale_build/config.py
@@ -47,6 +47,7 @@ TRAIN = get_env_variable('TRUENAS_TRAIN', str)
 TRUENAS_BRANCH_OVERRIDE = get_env_variable('TRUENAS_BRANCH_OVERRIDE', str)
 TRY_BRANCH_OVERRIDE = get_env_variable('TRY_BRANCH_OVERRIDE', str)
 VERSION = get_env_variable('TRUENAS_VERSION', str, f'{_VERS}-{BUILD_TIME_OBJ.strftime("%Y%m%d-%H%M%S")}')
+TRUENAS_VENDOR = get_env_variable('TRUENAS_VENDOR', str)
 
 
 # We will get branch overrides and identity file path overrides from here
--- a/scale_build/image/iso.py
+++ b/scale_build/image/iso.py
@@ -5,6 +5,7 @@ import shutil
 import tarfile
 import tempfile
 import time
+import json
 
 import requests
 
@@ -12,6 +13,7 @@ from scale_build.exceptions import CallError
 from scale_build.utils.manifest import get_manifest
 from scale_build.utils.run import run
 from scale_build.utils.paths import CD_DIR, CD_FILES_DIR, CHROOT_BASEDIR, CONF_GRUB, PKG_DIR, RELEASE_DIR, TMP_DIR
+from scale_build.config import TRUENAS_VENDOR
 
 from .bootstrap import umount_chroot_basedir
 from .manifest import get_image_version, update_file_path
@@ -53,6 +55,11 @@ def make_iso_file():
     # Set /etc/hostname so that hostname of builder is not advertised
     with open(os.path.join(CHROOT_BASEDIR, 'etc/hostname'), 'w') as f:
         f.write('truenas-installer.local')
+
+    os.makedirs(os.path.join(CHROOT_BASEDIR, 'data'))
+    if TRUENAS_VENDOR:
+        with open(os.path.join(CHROOT_BASEDIR, 'data/.vendor'), 'w') as f:
+            f.write(json.dumps({'name': TRUENAS_VENDOR}))
 
     # Copy the CD files
     run(f'rsync -aKv {CD_FILES_DIR}/ {CHROOT_BASEDIR}/', shell=True)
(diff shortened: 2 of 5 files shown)


aiden3c:
@themylogin to answer your question on the old PR about the keyword argument, it's just to be more explicit with what's being passed over. Even if it's the only argument.

scale_build/image/iso.py
```diff
@@ -5,14 +5,17 @@
...
+import json
 
 import requests
 
 from scale_build.exceptions import CallError
 from scale_build.utils.manifest import get_manifest
 from scale_build.utils.run import run
 from scale_build.utils.paths import CD_DIR, CD_FILES_DIR, CHROOT_BASEDIR, CONF_GRUB, PKG_DIR, RELEASE_DIR, TMP_DIR
-
+from scale_build.config import (
```
yocalebo, line 16:
```suggestion
from scale_build.config import TRUENAS_VENDOR
```

scale_build/iso.py
```diff
@@ -9,6 +9,9 @@
 from .image.manifest import get_image_version, update_file_path
 from .utils.logger import LoggingContext
 from .utils.paths import LOG_DIR, RELEASE_DIR
+from .config import (
```
yocalebo, line 12:
```suggestion
from .config import TRUENAS_VENDOR
```

yocalebo (changes requested):
Looks good, just don't like how the TRUENAS_VENDOR has been imported :smile:

scale_build/image/manifest.py
```diff
@@ -64,12 +64,13 @@ def build_release_manifest(update_file, update_file_checksum):
...
 
 
-def get_image_version():
+def get_image_version(vendor=None):
     if not os.path.exists(RELEASE_MANIFEST):
         raise CallError(f'{RELEASE_MANIFEST!r} does not exist')
 
     with open(RELEASE_MANIFEST) as f:
-        return json.load(f)['version']
+        vendor = f"-{vendor}" if vendor else ""
```
sonicaj, line 72: Can we please be consistent with quotes usage ? i.e if file is using single quotes, let's use single quotes and double quotes otherwise

themylogin:
> @themylogin to answer your question on the old PR about the keyword argument, it's just to be more explicit with what's being passed over. Even if it's the only argument.

Do we have some agreement about passing everything as a keyword argument?

yocalebo:
> > @themylogin to answer your question on the old PR about the keyword argument, it's just to be more explicit with what's being passed over. Even if it's the only argument.
> 
> Do we have some sort of agreement about passing everything as a keyword argument?

There is nothing inherently wrong calling a function using a keyword argument when the functions definition uses keyword definitions? I'm not sure what the issue is with how this has been done.

scale_build/config.py
```diff
@@ -47,6 +47,7 @@ def get_normalized_value(value, _type, default_value=None):
 TRUENAS_BRANCH_OVERRIDE = get_env_variable('TRUENAS_BRANCH_OVERRIDE', str)
 TRY_BRANCH_OVERRIDE = get_env_variable('TRY_BRANCH_OVERRIDE', str)
 VERSION = get_env_variable('TRUENAS_VERSION', str, f'{_VERS}-{BUILD_TIME_OBJ.strftime("%Y%m%d-%H%M%S")}')
+TRUENAS_VENDOR = get_env_variable('TRUENAS_VENDOR', str)
```
sonicaj, line 50: Should we default to some value here ?
Motivation is that i think we should be consistent that the vendor file exists at all times

yocalebo: No, the vendor file should _NOT_ exist at all times. It only exists when someone builds an ISO image specifying this environment variable.

truenas_install/__main__.py
```diff
@@ -533,6 +534,9 @@ def main():
                     write_progress(0.58, "Configuring serial port")
                     configure_serial_port(root, os.path.join(root, "data/freenas-v1.db"))
 
+                if TRUENAS_VENDOR:
```
sonicaj, line 537: I doubt this would work, this script is executed when the image is being installed and at that time it won't have access to this var. Have we tested this ?

themylogin:
> > > @themylogin to answer your question on the old PR about the keyword argument, it's just to be more explicit with what's being passed over. Even if it's the only argument.
> > 
> > 
> > Do we have some sort of agreement about passing everything as a keyword argument?
> 
> There is nothing inherently wrong calling a function using a keyword argument when the functions definition uses keyword definitions? I'm not sure what the issue is with how this has been done.

Nothing wrong, just excessively verbose `vendor=VENDOR`

truenas_install/__main__.py
themylogin: Also this should be done on Line 467 where other files in `{root}/data` are being created on fresh install. My suggestion is to just copy `/data` from the original system ane make this script agnostic of `.vendor` or other files we might want to copy from root in the future.

aiden3c:
Tested and working on a local VM, running a build on Jenkin's right now

themylogin (approved):
Excellent!

yocalebo (changes requested):
please squash these into 1 commit before merging

yocalebo (approved):
Thanks!

PR #666, opened 2024-07-12
NAS-130002 / 24.10 / Show vendor name during install

aiden3c:
This injects the vendor name into our grub.cfg for the installer grub entries

### DIFF
--- a/scale_build/image/iso.py
+++ b/scale_build/image/iso.py
@@ -37,8 +37,7 @@ def install_iso_packages_impl():
     # Inject vendor name into grub.cfg
     with open(CONF_GRUB, 'r') as f:
         grub_cfg = f.read()
-    if TRUENAS_VENDOR:
-        grub_cfg = grub_cfg.replace('TrueNAS SCALE', TRUENAS_VENDOR)
+    grub_cfg = grub_cfg.replace('$vendor', TRUENAS_VENDOR or 'TrueNAS SCALE')
 
     os.makedirs(os.path.join(CHROOT_BASEDIR, 'boot/grub'), exist_ok=True)
     with open(os.path.join(CHROOT_BASEDIR, 'boot/grub/grub.cfg'), 'w') as f:
--- a/scale_build/image/iso.py
+++ b/scale_build/image/iso.py
@@ -34,8 +34,15 @@ def install_iso_packages_impl():
     for package in get_manifest()['iso-packages']:
         run_in_chroot(['apt', 'install', '-y', package])
 
+    # Inject vendor name into grub.cfg
+    with open(CONF_GRUB, 'r') as f:
+        grub_cfg = f.read()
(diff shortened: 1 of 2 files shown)


scale_build/image/iso.py
```diff
@@ -34,8 +34,15 @@ def install_iso_packages_impl():
     for package in get_manifest()['iso-packages']:
         run_in_chroot(['apt', 'install', '-y', package])
 
+    # Inject vendor name into grub.cfg
+    with open(CONF_GRUB, 'r') as f:
+        grub_cfg = f.read()
+    if TRUENAS_VENDOR:
+        grub_cfg = grub_cfg.replace('TrueNAS SCALE', TRUENAS_VENDOR)
```
themylogin, line 41: Too broad, we can end up with unexpected substitutions. Can we put `TRUENAS_VENDOR` during the config file generation?

themylogin (changes requested):

scale_build/image/iso.py
aiden3c: The config under `boot/grub/grub.cfg` is copied over in this function anyway, I'm just modifying it if relevant.

themylogin: Can we modify default `grub.cfg` to have something like `$vendor` here?

aiden3c: I'm not sure how I'd go about setting up that variable in the first place, but if we can then that would be the better solution. Later in the file there's some grub.cfg injection similar to what I'm doing (with the string replace and all), so I modeled my changes after that

themylogin: Look what `CONF_GRUB` variable is, it points to a file in this repository

aiden3c: I mean how I'd set it within the .cfg itself. I think I'd need to check for the .vendor file, parse the JSON, then use that. If that's the intention then I will do that

themylogin: You just edit cfg file in the repository and replace `Start TrueNAS SCALE Installation` with `Start $vendor Installation` so that we can reliably replace `$vendor` later

aiden3c: With those changes, where would we set that variable so it's picked up in the `grub.cfg`? Not super familiar with the grub config process, especially with things like those external variables.

themylogin: It will not be picked up, you'll have to replace it the same way yo do now

aiden3c: Ohhhhhh okay I see what you're suggesting now.

So instead of looking for `TrueNAS SCALE` and doing the substring replacement with that, we instead have `$vendor` as a placeholder and we look for that and replace accordingly. Defaulting to TrueNAS SCALE.

themylogin (approved):

PR #675, opened 2024-08-01
NAS-130356 / 24.10 / `intel-gpu-tools`

creatorcary:
[NAS-130356](https://ixsystems.atlassian.net/browse/NAS-130356): Adding `intel-gpu-tools` package at community request.

Passed all tests except `test_system_general_ui_rollback.test_system_general_ui_rollback`.
http://jenkins.eng.ixsystems.net:8080/job/master/job/api_tests/1760/

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -308,6 +308,9 @@ additional-packages:
 - name: systemd-container
   comment: requested by community (NAS-123533)
   install_recommends: false
+- name: intel-gpu-tools
+  comment: requested by community (NAS-130356)
+  install_recommends: false
 
 #
 # List of additional packages installed into TrueNAS SCALE ISO file


aiden3c (approved):

scale_build/clean.py
```diff
@@ -25,6 +25,10 @@ def clean_packages():
             os.makedirs(path)
 
 
+def clean_diversions():
+    shutil.rmtree(DIVERT_DIR, ignore_errors=True)
```
creatorcary, line 29: I think this is failing silently. Will have to investigate.

yocalebo (changes requested):
Please don't combine tickets. Each ticket needs to be kept separate from one another especially since they're fixing unrelated issues.

This allows us to more easily do git blame, git log and git revert in the event we need to revert something.

yocalebo (approved):

PR #676, opened 2024-08-01
NAS-130351 / 24.10 / clean `/var/trash`

creatorcary:
[NAS-130351](https://ixsystems.atlassian.net/browse/NAS-130351): Cleaning out /var/trash package diversions.

API tests: http://jenkins.eng.ixsystems.net:8080/job/master/job/api_tests/1769/


scale_build/clean.py
```diff
@@ -25,6 +25,10 @@ def clean_packages():
             os.makedirs(path)
 
 
+def clean_diversions():
+    shutil.rmtree(DIVERT_DIR, ignore_errors=True)
```
sonicaj, line 29: You are nuking builder's `/var/trash` here, i assume you want this for the scale image itself ? In that case we need to fix it's path below

creatorcary: Thanks again

bmeagherix (approved):

PR #677, opened 2024-08-03
NAS-130395 / 24.10 / Add nscd to base build

anodos325:
This was originally installed as a recommended package for nslcd, and so
it must be re-added now that nslcd is no longer included in base packages.


yocalebo (approved):
Do we need "install_recommends: true" for this package? Don't even know if it brings anything else in as extra.

PR #693, opened 2024-08-15
Add incus as a dependency

william-gr:
Adding incus as a dependency to facilitate testing/development using nightly images.


yocalebo (approved):
Nice!

william-gr:
@yocalebo apparently it will need a couple groups, is this ok? Will it need some changes in migration?

william-gr:
retest this please

william-gr:
apparmor and dnsmasq-base are already part of the system.

incus-agent is not important for us at this stage.

PR #696, opened 2024-08-26
TEN-1474 Option added to prevent deletion of ISO files during Make

alexandrabain:
Update to allow both standard and vendor ISO to be produced from a pipeline without the iso which is generated first being deleted.

### DIFF
--- a/scale_build/image/iso.py
+++ b/scale_build/image/iso.py
@@ -14,6 +14,8 @@ from scale_build.utils.manifest import get_manifest
 from scale_build.utils.run import run
 from scale_build.utils.paths import CD_DIR, CD_FILES_DIR, CHROOT_BASEDIR, CONF_GRUB, PKG_DIR, RELEASE_DIR, TMP_DIR
 from scale_build.config import TRUENAS_VENDOR
+from scale_build.config import PRESERVE_ISO
+
 
 from .bootstrap import umount_chroot_basedir
 from .manifest import get_image_version, update_file_path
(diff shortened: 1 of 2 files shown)


scale_build/image/iso.py
```diff
@@ -14,6 +14,8 @@
 from scale_build.utils.run import run
 from scale_build.utils.paths import CD_DIR, CD_FILES_DIR, CHROOT_BASEDIR, CONF_GRUB, PKG_DIR, RELEASE_DIR, TMP_DIR
 from scale_build.config import TRUENAS_VENDOR
+from scale_build.config import PRESERVE_ISO
+
```
yocalebo, line 18: Seems like an extra newline was added.

scale_build/image/iso.py
```diff
@@ -45,8 +47,9 @@ def install_iso_packages_impl():
 
 
 def make_iso_file():
-    for f in glob.glob(os.path.join(RELEASE_DIR, '*.iso*')):
-        os.unlink(f)
+    if PRESERVE_ISO:
```
yocalebo, line 50: The logic is inverted. If this is set to true, it will remove the ISOs but I think the exact opposite is wanted?

scale_build/image/iso.py
alexandrabain: Removed

scale_build/image/iso.py
alexandrabain: Fixed

kmoore134 (approved):

PR #741, opened 2024-10-14
NAS-131773 / 25.04 / Fix unsquashfs error reporting

themylogin:
Types mismatch results in installer crashing when unsquashfs error occurs


themylogin:
time 0:15

yocalebo (approved):

PR #772, opened 2024-11-22
NAS-132388 / 25.04 / Save snmp config directory in upgrades

mgrimesix:
SNMP uses /var/lib/snmp to hold it's configuration, including v3 user information.
This directory should be preserved across an upgrade.

This PR adds preservation of the SNMP configuration directory during an upgrade.
This same fix is valid for backport to 24.10.2.


themylogin (approved):

yocalebo:
This is trivial enough to back port to 24.10.1 and we're early enough in test cycle to get it in before we release. I've gone ahead and added the 24.10.1 back port label.

PR #787, opened 2024-12-09
NAS-132964 / 25.04 / Add audit rules to build

anodos325:
This commit adds the audit rules package to the build and adds a step in which we generate the privileged rules set based on the update file before generating mtree.


yocalebo:
Should we add to mtree?

yocalebo (approved):

anodos325:
> Should we add to mtree?

It should get automatically added to the mtree results.

anodos325:
> > Should we add to mtree?
> 
> It should get automatically added to the mtree results.

And it didn't so I had to expand PR :)

PR #789, opened 2024-12-09
NAS-132966 / 25.04 / Always add python3 as an explicit dependency

anodos325:
It's somewhat challenging to predict what packages may have subtle errors due to us having different python version that what is provided by upstream. This commit ensures that when we're building packages we always have our own python package present.


yocalebo (approved):

sonicaj (approved):
We should set batch priority for python3 in build manifest as well, LGTM otherwise

PR #812, opened 2025-01-23
NAS-133889 / 25.10 / Add usbutils

william-gr:
https://forums.truenas.com/t/please-add-lsusb-all-ls-commands-to-scale/9099

### DIFF
--- a/conf/build.manifest
+++ b/conf/build.manifest
@@ -142,6 +142,9 @@ base-packages:
 - name: lsscsi
   # requested by the support team (NAS-130189)
   install_recommends: false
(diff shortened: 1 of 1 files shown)


william-gr:
@yocalebo @kmoore134 ping

kmoore134 (approved):

kmoore134:
Ideally we get this into the product, since it is incredibly useful for troubleshooting and device discovery in general when doing pass-through, etc. Package is pretty benign and tiny <400K, so should be super low risk.

yocalebo (approved):

PR #824, opened 2025-02-12
NAS-133968 / 25.10 / Update mtree to better support cryptographic verification for STIG

mgrimesix:
Update mtree to support cryptographic verification for STIG
-------------------------------------------------------------------
Our goal is to have zero discrepancies at install.

- Added a few files to the `delete` list and `ignore` list.  These were discovered during testing.
- Created a list of files that require `chmod`

Explanation of the files added to delete:
- `etc/audit/rules.d/audit.rules`  is not used by TrueNAS.  This file is deleted by the audit subsystem.
- `etc/rc[2345].d/K01ssh` are symlinks used by the legacy init.d.  The systemd subsystem deletes these on ssh start.

### DIFF
--- a/scale_build/image/mtree.py
+++ b/scale_build/image/mtree.py
@@ -45,7 +45,7 @@ ETC_FILES_TO_REMOVE = [
     'etc/rc3.d/K01ssh',
     'etc/rc4.d/K01ssh',
     'etc/rc5.d/K01ssh',
-    'etc/initramfs-tools/modules',  # These three are not used by systemd
+    'etc/initramfs-tools/modules',  # These two are not used by systemd
     'etc/modules',
 ]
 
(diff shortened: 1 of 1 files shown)


scale_build/image/mtree.py
```diff
@@ -40,6 +41,17 @@
...
+    'etc/rc3.d/K01ssh',
+    'etc/rc4.d/K01ssh',
+    'etc/rc5.d/K01ssh',
+]
+
+# Some files or directories get the permission mode changed on install.
+# The following is a list of tuples (files, mode).
+# Preemptively change the mode before generating the mtree.
+OBJS_TO_FIXUP = [
+    ('var/lib/incus', 0o744)
```
anodos325, line 54: I think this is a bug we need to fix. 744 doesn't make sense.

mgrimesix: fixed at source and removed entry.  I'll leave the mode fix up infrastructure, unless you object

anodos325 (approved):

PR #842, opened 2025-04-16
Allow specifying secret env variables in the build

sonicaj:
This PR adds ability to specify secret env variables which can be passed to `debuild` when the debian package builds.


scale_build/utils/manifest.py
```diff
@@ -198,6 +199,21 @@
 }
 
 
+@functools.cache
+def get_secret_env():
+    try:
+        with open(SECRETS_FILE, 'r') as f:
+            secrets = yaml.safe_load(f.read())
```
yocalebo, line 206: Do we really have to use yaml? Why not json?

scale_build/packages/build.py
```diff
@@ -160,7 +171,10 @@ def build_command(self):
             return self.buildcmd
         else:
             build_env = f'DEB_BUILD_OPTIONS={self.deoptions} ' if self.deoptions else ''
-            return [f'{build_env} debuild {" ".join(self.deflags)}']
+            env_flags = [
+                f'-e{k}={shlex.quote(v)}' for k, v in self._get_chroot_env().items()
```
themylogin, line 175: How secret are these secrets? Is it a problem that these will be visible in `ps ax` output? Or does debuildhide them?

themylogin (approved):

billohanlon:
This is general purpose?  Or is it for sending tracebacks to sentry?  Or both?

sonicaj:
@billohanlon the feature is general purpose, although the use case we have right now is only for integrating sentry in webui

scale_build/packages/build.py
sonicaj: It shows up like this now
```
Running build command: ' debuild -eRELEASE_VERSION -eSENTRY_AUTH_TOKEN --no-lintian -j32 -us -uc -b'
```

yocalebo (approved):

PR #853, opened 2025-05-06
NAS-135631 / 25.10 / Update build to pull in GE mirrors and use bookworm-backports as default

sonicaj:


scripts/grub.cfg
```diff
@@ -71,12 +71,13 @@ play 960 440 1 0 4 440 1
 menuentry --hotkey=i --id=video 'Start $vendor Installation' {
     load_video
     set background_color=black
-    linux    /vmlinuz gfxpayload=text quiet nomodeset boot=live toram=filesystem.squashfs $gfx_console
+    linux    /vmlinuz gfxpayload=text quiet boot=live toram=filesystem.squashfs $gfx_console
```
yocalebo, line 74: Why are we removing nomodeset?

sonicaj: It is a change we do not want to merge and was one of the things done while debugging issues related to grub - testing without this (change has been reverted)

yocalebo (approved):

PR #857, opened 2025-05-13
NAS-135854 / 25.10 / Add python version check to truenas_install module

anodos325:
This commit adds a basic python version check to the truenas_install module so that we can raise a useful error message if someone is somehow running the installer in the context of a really old python version (like in 13).


anodos325:
Sample of 13.0-U6.7 system with revision

image

yocalebo (approved):

PR #858, opened 2025-05-14
NAS-135869 / 25.10 / Set image default update profile

themylogin:


themylogin:
http://jenkins.eng.ixsystems.net:8080/job/master/job/custom/1113/

themylogin:
Not necessary. The default update profile is `null` as per last meeting.

PR #874, opened 2025-06-13
`TRUENAS_SUPPORTED` environment variable

themylogin:


yocalebo (changes requested):
I'd rather change this to "experimental". So `TRUENAS_EXPERIMENTAL_BUILD` as the env var and `experimental_build` as the key in the manifest.

themylogin:
@yocalebo do we change the API to use `experimental` as well?

yocalebo:
> @yocalebo do we change the API to use `experimental` as well?

yeah

PR #910, opened 2025-08-22
NAS-133371: Add WebShare service support to scale-build

kmoore134:
- Add truenas-file-manager and truesearch to build manifest
- Update package installation to handle external packages from assets


yocalebo:
ping @truenas/middleware introduces a simplified mechanism for downloading packages during build.

scale_build/image/update.py
```diff
@@ -163,9 +163,144 @@ def post_rootfs_setup():
...
+        except Exception as e:
+            logger.error(f'Failed to install {package_name}: {e}')
+            raise RuntimeError(f'Failed to install {package_name} package: {e}')
+        finally:
+            # Clean up the package from chroot tmp
+            if os.path.exists(chroot_tmp_path):
+                os.unlink(chroot_tmp_path)
+
+
+def install_truenas_file_manager():
```
themylogin, line 226: Should we just put deb files URL in the config? These three methods don't seem to be different from one another. We can then just install all packages from `manifest["external-packages"]` without the hardcoding their names in the code.

themylogin (approved):

anodos325 (approved):

PR #933, opened 2025-09-24
NAS-137634 / 26.04 / Install restic from Debian

creatorcary:
This replaces our custom package at https://github.com/truenas/restic and upgrades to restic 0.18.0.


yocalebo (changes requested):
We don't need the recommended packages for this program.

yocalebo (approved):

PR #937, opened 2025-10-01
NAS-137831 / 26.04 / Remove /etc/dhcpcd.conf from mtree

bmeagherix:
In a recent middleware PR (#[17193](https://github.com/truenas/middleware/pull/17193)) `dhcpcd.conf.mako` was added, therefore we should no longer check it.


bmeagherix:
time 0:10

PR #1002, opened 2026-07-29
NAS-142008 / 25.10.6 / Bump nvidia to 580.173.02 (LTSB)

ixhamza:
Required for https://github.com/truenas/linux/pull/325.

570.172.08 does not build against recent 6.12 kernels. v6.12.97 picked up [`0d63176d355e`](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=0d63176d355e) (backport of mainline [`337b1b566db0`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=337b1b566db0)), which added an `exclude_bars` argument to `pci_resize_resource()`:

```
nv-pci.c:240:9: error: too few arguments to function 'pci_resize_resource'; expected 4, have 3
```

**Why bump rather than patch:**

- 570 is a Production Branch that went [EOL in January 2026](https://endoflife.date/nvidia), so it never picked up the fix and no 570.x release will build against this kernel.
- [Patching 570 is trivial](https://github.com/truenas/scale-build/commit/92a731306d0e532df0addc7e62899e481620149f) in itself, but it only covers this one break. 570 is well behind on kernel API adaptations, so the next stable bump could need another patch.
- 580 is the current LTSB, supported until June 2028, and [580.173.02](https://github.com/NVIDIA/open-gpu-kernel-modules/releases/tag/580.173.02) is its latest release. It carries NVIDIA's fix, shipped in 580.126.09.
- 580 is also [the last branch with full Maxwell/Pascal/Volta support](https://www.phoronix.com/news/NVIDIA-580-Linux-Driver-Last-HW), so older cards used for transcoding stay supported.

### Testing

Built both drivers against the production headers and `Module.symvers` pulled off a running v6.12.99 VM: 570.172.08 fails as above, 580.173.02 builds clean (214 objects, all 5 modules).

On the VM itself the sysext merges, the module loads with no unresolved symbols, and the userspace lib version matches the kernel module.

> **Note:** the test VMs have no GPU attached, so nvidia-smi enumeration, CUDA, NVENC/NVDEC and container passthrough are untested. That needs bare metal or a VM with GPU passthrough.


yocalebo (approved):

ixhamza:
time 04:00

